PR #771, opened 2019-10-09
INFRA-439 Fix how CSVs are made to take into account different quote …

dallemang:
…protocols

This also comments out some software we are not currently using (latex, bibtex), and updates the version of Jena to be current.


trypuz:
> Changes in infrastructure (not in FIBO content) appear in the ontology-publisher repository in github. But once that has been tested, that makes a new version of the docker image, which is indicated by an increase in the version number. The way to run that for FIBO is to change the corresponding number in the Jenkinsfile, so that it will call the new docker.
> 
> To make this process work properly, we need to couple the update process here in fibo with the process in ontology-publisher.

@dallemang is there anything to review in this PR?

rivettp:
Is this process documented anywhere apart from this PR?

dallemang:
I don't think so - I always have to re-think it each time to remember how it works.    We really should sort out this workflow, especially how we update the docker version number in master once we decide that we like the change.  So for example, I have no idea whether the changes for this three-month old PR have actually made it in to the current publish.

dallemang:
I want to keep this discussion, but get the PR off the queue, so I am closing this without merge.

Issue #97, opened 2014-06-10
Incorrect definition of Organization


MikeHypercube:
The definition given in FND/ORG/ORG for Organization is: 

"A social unit of people, systematically structured and managed to meet a need or pursue collective goals on a continuing basis."
Source is given as: http://www.BusinessDic-tionary.com/

This is incorrect: the intended scope for organization, and the properties modeled for it, are that an organization may have members which are any kind of autonomous agent. The definition limits membership of an organization to "people", which on any interpretation means human beings alone. 

Proposed resolution: 
Change definition to: 
"An organization consists of one, two or more individuals participating in an entity which is an autonomous thing, usually set up for some purpose. "
Change definitionOrigin to: EDM Council

Note: I was going to propose the W3C Organization class definition, but this also has "people". This may have (separate) implications for alignment with the W3C Organization ontology.

rivettp:
Reopening this since the current definition, though quite different, is still specific to persons:
"a unique framework of authority within which a person or persons act, or are designated to act, towards some purpose, such as to meet a need or pursue collective goals on a continuing basis"

The new definition additionally seems over-restrictive and does not encompass an InformalOrganization which is a subclass.

ElisaKendall:
This definition was taken from the ISO 6523 standard, which could indeed include informal organization.  

Every organization that we have needed to address based on FIBO use cases must have at least one person as its documented head, at a minimum on paper, including trusts, funds, and all other organizations that we care about.  We have no use cases for informal organizations that are not incorporated, even a community soccer team needs to have incorporated as some kind of benefit corporation in order to open a bank account in the US or EU.  If you have a use case for an organization that is not banked and is more informal than a PTA or soccer team, then post it.  Otherwise, using the ISO definition of organization is at least unimpeachable from a source perspective.

I propose that we close this because the notion of an informal organization is not needed for any use case we have at present.

rivettp:
In that case we should deprecate InformalOrganization because currently we're asserting that it complies with that pretty heavyweight definition.

MikeHypercube:
Indeed. The original use case, established during SME reviews, was the need to talk about cartels and other informal groupings, including those that are organizations as well as those that are not. 

Separately, it should be clear that the meaning of a concept does not change according to the scope of ways we happen to know it might be used. Either we define 'organization' as what it actually is (leaving out those sub classes we don't think we need), or we define something that corresponds to Elisa's narrower scope, but then it is not 'organization' as the concept is most broadly defined.

PR #830, opened 2020-01-20
FND-276 #comment - GitHub issue #97

ElisaKendall:
## Description

Revise the definition of organization as suggested, to replace 'person' with 'individual' and eliminate unnecessary / unused legitimate organizations ontology to simplify the hierarchy

Fixes: #97, JIRA issue BE-276

## Checklist:


ElisaKendall:
How would you like it to read?  The only thing I changed in the definition itself was to change "person or persons" to individual or individuals.  The rest was there before.

ElisaKendall:
I would rather quote a standard definition for organization than invent yet another one.

trypuz:
@ElisaKendall; [The sign-off is missing.](https://github.com/edmcouncil/fibo/pull/830/checks?check_run_id=401645540)

rivettp:
The definition I already quoted from the W3C standard, which we claim to be aligned with, is more general and more appropriate for FIBO than the narrow, authority-based one used by ISO for one specific standard.

ElisaKendall:
The W3C definition is circular whereas the ISO definition is not.  I could remove "of authority" but I much prefer referencing an ISO definition that meets ISO 704 requirements, which we have said that we are doing.  Mike objected to limiting organizations to people - so I'm not sure what we are talking about in any case. All of the uses of organization that I know of in FIBO are of groups of one or more people.  I don't believe we have said that we would cover organizing artificial agents with our definition of organization, thus I propose the following:

unique framework within which a person or persons act, or are designated to act, towards some purpose

At least is is not circular, and eliminates the issue with "of authority", which I disagree is a problem, since all of our organizations are academic, research institutes, companies, or government entities.

rivettp:
Defining Organization as a "unique framework" seems pretty arcane to any business person. And "framework" itself is used in all sorts of different ways. 
Note that the examples of organizations you listed would be FormalOrganizations in FIBO: the generic Organization is intended to encompass social clubs etc as clearly stated, even though we might not need them for FIBO per se.
If you think "organized" is too circular in the W3 definition (I don't) then how about "grouped"?

trypuz:
> @ElisaKendall; [The sign-off is missing.](https://github.com/edmcouncil/fibo/pull/830/checks?check_run_id=401645540)

DCO is OK now.

Issue #865, opened 2020-02-11
fibo-sec-sec-iss:SecuritiesOffering in SEC and BP


trypuz:
[fibo-sec-sec-iss:SecuritiesOffering](https://spec.edmcouncil.org/fibo/ontology/SEC/Securities/SecuritiesIssuance/SecuritiesOffering) is defined both in [SecuritiesIssuance.rdf](https://github.com/edmcouncil/fibo/blob/master/SEC/Securities/SecuritiesIssuance.rdf) and [IssuanceProcessTerms.rdf](https://github.com/edmcouncil/fibo/blob/master/BP/SecuritiesIssuance/IssuanceProcessTerms.rdf) with different restrictions and annotations. The SEC  specification seems to be the correct one.

rivettp:
There is nothing wrong with one ontology adding further restrictions on a concept defined in another, since the concepts needed might well not be available in the original or its set of imported ontologies (bearing in mind our rule to avoid import cycles)
That is the case here.
However what are wrong are some of the annotation properties such as the label and definition in IssuanceProcessTerms.

ElisaKendall:
Issuance Process Terms has lots of challenges - this is a very provisional ontology that has not been reviewed.  I'll take a crack at it to eliminate some of the duplication, though, including this class.

PR #882, opened 2020-02-14
BP-865 #comment - eliminated redundant concepts including SecuritiesO…

ElisaKendall:
…ffering from provisional issuance ontologies

## Description

This resolution eliminates a number of redundant concepts from the [IssuanceProcessTerms](https://spec.edmcouncil.org/fibo/ontology/BP/SecuritiesIssuance/IssuanceProcessTerms/) and related provisional ontologies.

Fixes: #865

## Checklist:


BP/SecuritiesIssuance/IssuanceProcessTerms.rdf
```diff
@@ -146,7 +146,7 @@
 		<rdfs:subClassOf rdf:resource="&fibo-fnd-dt-fd;RegularSchedule"/>
 		<rdfs:subClassOf>
 			<owl:Restriction>
-				<owl:onProperty rdf:resource="&fibo-bp-iss-trm;expression"/>
+				<owl:onProperty rdf:resource="&fibo-fnd-acc-cur;hasMonetaryAmount"/>
```
rivettp: This does not seem right - the domain is MonetaryAmount not MonetaryMeasure.

ElisaKendall: I think you are right, but I also think the restriction is unnecessary.  There is an inherited restriction for the payment which includes an amount.  That can be set via an algorithm, on an ad hoc basis, or as a regularly scheduled set of payments.  The question here is whether or not the class itself is needed at all.  It seems redundant with its parent class - whose definition is a schedule for partial payments.  I'm inclined to delete the whole class, including the restriction.  What do you think?

ElisaKendall: Nothing references it, FYI

rivettp: The class doesn't seem useful to me, but Ill approve the PR as it is, now you've made the change.

Issue #814, opened 2020-01-09
The AboutFIBOProd and AboutFIBODev ontologies have bogus IRIs and need to be documented with metadata that can help github users


ElisaKendall:
These files have only been used by "insiders" while working with the ontologies in github to date, but with an open repository they will become essential to developers. They need to be documented, including relevant header details, just as any other ontology in FIBO and managed appropriately. They also need to be kept up to date as part of our process.

These files work with the github versions of the ontologies, rather than the published versions, whose IRIs are distinct from these. They were automatically published at some point but have been corrupted over time. The process that used to generate them should either be repaired or abandoned.

ElisaKendall:
This issue corresponds to FND-273 in JIRA.

trypuz:
@dallemang , @ElisaKendall , what exactly should be done to fix this issue?

ElisaKendall:
@trypuz - I've put this on hold until I finish merging some of the loans ontologies, because they need to be fixed manually.  There should be a nice header and information about each of the versions, like with the all files I've been maintaining under each domain area, but I don't want to do this until after I've eliminated / merged more ontologies.

PR #962, opened 2020-03-26
FND_295 - The AboutFIBOProd and AboutFIBODev ontologies have bogus IRIs and need to be documented with metadata that can help github users

ElisaKendall:

## Description

Revised top-level, github versions of the FIBO about files to provide proper header metadata, a proper IRI for each, and organize the ontologies so that they are grouped by domain area for ease of use by FIBO developers/github users.

Fixes: #814  / FND-295

## Checklist:

### DIFF
--- a/AboutFIBODev.rdf
+++ b/AboutFIBODev.rdf
@@ -1,318 +1,408 @@
-<rdf:RDF
-    xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
-    xmlns:owl="http://www.w3.org/2002/07/owl#"
-    xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#"
-    xmlns:xsd="http://www.w3.org/2001/XMLSchema#">
-  <owl:Ontology rdf:about="file:///AboutFIBO">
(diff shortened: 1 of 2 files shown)


ElisaKendall:
So the challenge is that people that are using FIBO as a reference, using Protege, and having access to github, but that are not programmers and don't have a clue as to how to generate load files -should- be able to load everything easily, and in a way that gives them a bit of metadata about what they have loaded. 

I use these files constantly to check to make sure that pellet runs on everything I check in.  I use the all files at the domain level and I use the all prod file at the top level as part of my QA process.  I expect every contributor to do the same.  If you make a change you are committing to having run at least one reasoner, possibly more than one, to ensure that the changes you are making are logically consistent.  To do that you either have to have your environment set up so that you can run various reasoners and rule engines on the command line, or you run them via tools like protege, or you load everything into a triple store and run pellet inside of stardog, which is the only one I'm aware of with built-in reasoning.  If we don't do that, we are letting our community down.

So I need these files and so does anyone else who contributes, and no one has fixed the logic in the tools that are supposedly generating them to make the files that are produced reliably correct.  And they are not run at all automatically anymore as far as I can tell.  But these load files are really helpful to people that are using them to review content for our FCT calls, like Citi is doing.  And, they should have decent metadata to let people know what they are loading.

Issue #835, opened 2020-01-22
IND does not allow representation of unemployment in millions


rivettp:
It only allows a percentage.
Not only are the numbers *published* with actual numbers, but that is frequently used in economic discussions.
For example from https://tradingeconomics.com/united-states/unemployment-rate "The US unemployment rate held steady at 3.5 percent in December 2019, remaining at the lowest level since 1969 and in line with market expectations. The number of unemployed people decreased by 58 thousand to 5.75 million while employment rose by 267 thousand to 158.80 million. "

The ontology does not allow you represent the 5.75 million.
This was previously discussed in issue #62 - it was requested there that a new issue be raised.

ElisaKendall:
The unemployment rate is defined as having an index value which is a ratio value; a ratio value has a number associated with it, not a percentage, so I'm confused about what's missing here.  At the higher level, all economic indicators have an indicator value that is a quantity value ...

rivettp:
We cannot represent the sentence I quoted "The number of unemployed people decreased by 58 thousand to 5.75 million while employment rose by 267 thousand to 158.80 million.". While not the "unemployment rate" it's important and distinct economic indicator.

ElisaKendall:
So, we can represent the 'unemployed population', that has a universe size in non-negative integer,  although maybe that should be non-negative number, or decimal, for 5.78 -- we don't have the magnitude in millions or thousands, though, and it should be of type decimal (quantity value rather than nonNegativeInteger is needed there.  But we don't have the decrease or increase as a comparative number in that intervening time ... is that what you are asking for?

rivettp:
If we have the unemployed population (also the employed population) of the US (or a State) with an associated dateTime that's indeed what I was looking for. People can calculate differences themselves.

ElisaKendall:
In the EconomicIndicators ontology, we have fibo-ind-ei-ei;UnemployedPopulation, which is a subclass of fibo-ind-ei-ei;CivilianLaborForce.  This class does not represent the complete picture, as there are many people that are no longer considered as part of the unemployed population because they are not counted as looking for work.  However, the class fibo-ind-ei-ei;UnemployedPopulation corresponds exactly to the number of unemployed people reported by the Bureau of Labor Statistics in the US.  

[...]

PR #1011, opened 2020-04-17
FND-298 - Several changes are needed to the definition of numeric index

ElisaKendall:

## Description

Revised the definition of numeric index to be numeric index value per discussion in the IND FCT; updated its use in economic indicators accordingly

Fixes: #835 / FND-298

## Checklist:

### DIFF
--- a/FND/Utilities/Analytics.rdf
+++ b/FND/Utilities/Analytics.rdf
@@ -769,9 +769,9 @@ For certain indices, one of the most common weighting factor is by market capita
 	</owl:ObjectProperty>
 	
 	<owl:ObjectProperty rdf:about="&fibo-fnd-utl-alx;hasRelativeComparativeDate">
-		<rdfs:subPropertyOf rdf:resource="&fibo-fnd-dt-fd;hasRelativeDate"/>
+		<rdfs:subPropertyOf rdf:resource="&fibo-fnd-dt-fd;hasDate"/>
 		<rdfs:label>has relative comparative date</rdfs:label>
-		<rdfs:range rdf:resource="&fibo-fnd-dt-fd;ExplicitDate"/>
+		<rdfs:range rdf:resource="&fibo-fnd-dt-fd;RelativeDate"/>
 		<skos:definition>specifies a date against which the value of a scoped measure is compared (e.g., one month prior, three months prior, etc., and typically against a prior release or average over prior releases)</skos:definition>
 	</owl:ObjectProperty>
 	
--- a/FND/Utilities/Analytics.rdf
+++ b/FND/Utilities/Analytics.rdf
@@ -401,14 +401,28 @@
 		</rdfs:subClassOf>
 		<rdfs:subClassOf>
 			<owl:Restriction>
-				<owl:onProperty rdf:resource="&fibo-fnd-utl-alx;hasBaseDate"/>
+				<owl:onProperty rdf:resource="&fibo-fnd-utl-alx;hasAnchorDate"/>
 				<owl:onClass rdf:resource="&fibo-fnd-dt-fd;ExplicitDate"/>
 				<owl:minQualifiedCardinality rdf:datatype="&xsd;nonNegativeInteger">0</owl:minQualifiedCardinality>
 			</owl:Restriction>
 		</rdfs:subClassOf>
 		<rdfs:subClassOf>
 			<owl:Restriction>
-				<owl:onProperty rdf:resource="&fibo-fnd-utl-alx;hasBasePeriod"/>
+				<owl:onProperty rdf:resource="&fibo-fnd-utl-alx;hasFixedComparativeDate"/>
+				<owl:onClass rdf:resource="&fibo-fnd-dt-fd;ExplicitDate"/>
+				<owl:minQualifiedCardinality rdf:datatype="&xsd;nonNegativeInteger">0</owl:minQualifiedCardinality>
+			</owl:Restriction>
+		</rdfs:subClassOf>
+		<rdfs:subClassOf>
+			<owl:Restriction>
+				<owl:onProperty rdf:resource="&fibo-fnd-utl-alx;hasRelativeComparativeDate"/>
+				<owl:onClass rdf:resource="&fibo-fnd-dt-fd;ExplicitDate"/>
+				<owl:minQualifiedCardinality rdf:datatype="&xsd;nonNegativeInteger">0</owl:minQualifiedCardinality>
+			</owl:Restriction>
+		</rdfs:subClassOf>
+		<rdfs:subClassOf>
+			<owl:Restriction>
+				<owl:onProperty rdf:resource="&fibo-fnd-utl-alx;hasRelativeComparativePeriod"/>
 				<owl:onClass rdf:resource="&fibo-fnd-dt-fd;ExplicitDatePeriod"/>
 				<owl:minQualifiedCardinality rdf:datatype="&xsd;nonNegativeInteger">0</owl:minQualifiedCardinality>
 			</owl:Restriction>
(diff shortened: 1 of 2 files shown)


FND/Utilities/Analytics.rdf
```diff
@@ -658,6 +674,13 @@ For certain indices, one of the most common weighting factor is by market capita
 		<fibo-fnd-utl-av:synonym>has independent variable</fibo-fnd-utl-av:synonym>
 	</owl:ObjectProperty>
 	
+	<owl:ObjectProperty rdf:about="&fibo-fnd-utl-alx;hasBaseDate">
+		<rdfs:subPropertyOf rdf:resource="&fibo-fnd-dt-fd;hasExplicitDate"/>
+		<rdfs:label>has base date</rdfs:label>
+		<rdfs:range rdf:resource="&fibo-fnd-dt-fd;ExplicitDate"/>
+		<skos:definition>specifies the base date against which the value of a numberic index for a more recent date is compared (or from which it stems)</skos:definition>
```
rivettp, line 681: typo "numberic"

FND/Utilities/Analytics.rdf
```diff
@@ -397,6 +399,20 @@
 				<owl:allValuesFrom rdf:resource="&fibo-fnd-dt-fd;RecurrenceInterval"/>
 			</owl:Restriction>
 		</rdfs:subClassOf>
+		<rdfs:subClassOf>
+			<owl:Restriction>
+				<owl:onProperty rdf:resource="&fibo-fnd-utl-alx;hasBaseDate"/>
```
rivettp, line 404: should have an explanatoryNote that only one of baseDate and basePeriod are expected to be used

FND/Utilities/Analytics.rdf
```diff
@@ -305,25 +308,24 @@
...
+	<owl:Class rdf:about="&fibo-fnd-utl-alx;NumericIndexValue">
+		<rdfs:subClassOf rdf:resource="&fibo-fnd-qt-qtu;QuantityValue"/>
 		<rdfs:subClassOf>
 			<owl:Restriction>
-				<owl:onProperty rdf:resource="&fibo-fnd-arr-doc;hasReportingPeriod"/>
-				<owl:onClass rdf:resource="&fibo-fnd-dt-fd;ExplicitDatePeriod"/>
-				<owl:qualifiedCardinality rdf:datatype="&xsd;nonNegativeInteger">1</owl:qualifiedCardinality>
+				<owl:onProperty rdf:resource="&fibo-fnd-utl-alx;isMeasureOf"/>
+				<owl:onClass rdf:resource="&fibo-fnd-utl-alx;ScopedMeasure"/>
+				<owl:minQualifiedCardinality rdf:datatype="&xsd;nonNegativeInteger">0</owl:minQualifiedCardinality>
```
rivettp, line 317: Why make this optional? When would this not be set - without the Measure then the value has no context/meaning

FND/Utilities/Analytics.rdf
```diff
@@ -738,6 +761,12 @@ For certain indices, one of the most common weighting factor is by market capita
 		<skos:definition>indicates the number of elements in a given population</skos:definition>
 	</owl:DatatypeProperty>
 	
+	<owl:DatatypeProperty rdf:about="&fibo-fnd-utl-alx;hasReferenceDateTime">
```
rivettp, line 764: not sure about "Reference" here which could get confused with "Base".

FND/Utilities/Analytics.rdf
```diff
@@ -738,6 +761,12 @@ For certain indices, one of the most common weighting factor is by market capita
...
+		<rdfs:label>has reference date and time</rdfs:label>
+		<rdfs:range rdf:resource="&fibo-fnd-dt-fd;CombinedDateTime"/>
+		<skos:definition>specifies a reference date and possibly a time for a given value</skos:definition>
```
rivettp, line 767: Circular definition. 
Also please avoid "specifies" in definitions. 
Need to clarify that's it's the time when the value was valid, as opposed to the time it was measured/sampled or published.
Finally I think this should be "the" not "a".

IND/EconomicIndicators/EconomicIndicators.rdf
```diff
@@ -389,7 +390,7 @@
 		<rdfs:subClassOf>
 			<owl:Restriction>
 				<owl:onProperty rdf:resource="&fibo-ind-ei-ei;hasIndicatorValue"/>
-				<owl:onClass rdf:resource="&fibo-fnd-utl-alx;NumericIndex"/>
+				<owl:onClass rdf:resource="&fibo-fnd-utl-alx;NumericIndexValue"/>
```
rivettp, line 393: This seems wrong! The CPI for the US, for example, has many values, each for a different date. Is this meant to be the latest value - in which case the property name should indicate that e.g. hasLatestValue. 
In general case I think the reference should be from the values to the index (which you have), not vice versa.

FND/Utilities/Analytics.rdf
ElisaKendall: fixed

FND/Utilities/Analytics.rdf
ElisaKendall: The FCT changed the names of these properties and added two additional properties to be clearer, but we need to review how they are applied.  hasBaseDate is now hasAnchorDate, which is the start for a series, to be clear about what that means.  We added hasFixedComparativeDate (e.g., the end of the last recession), hasRelativeComparativeDate (e.g. prior reported figure), and changed hasBasePeriod to hasRelativeComparativePeriod, again for clarity, but we need further analysis to determine whether they belong on scoped measure or something else.

FND/Utilities/Analytics.rdf
ElisaKendall: Still under discussion - it's possible that not all analyses that use scoped measures have an anchor date - maybe they do, and they certainly do for economic indicators, but scoped measure is broader than that.

FND/Utilities/Analytics.rdf
ElisaKendall: Not sure what you are thinking here - the name of the class this restriction is on is NumericIndexValue, whose definition is "change in some aggregate relative to the value of the aggregate as of some reference date" - the property definition does not say anything about when it was measured or published.  We could delete this altogether and just use hasObservedDateTime if you prefer.

IND/EconomicIndicators/EconomicIndicators.rdf
ElisaKendall: fixed

FND/Utilities/Analytics.rdf
ElisaKendall: See above on the renaming discussion.

FND/Utilities/Analytics.rdf
rivettp: My comment was about isMeasureOf not any date. When would a IndexValue not be linked to a Measure. 
BTW the property name seems wrong now I've looked at it again - it should be isValueOf not isMeasureOf.

FND/Utilities/Analytics.rdf
rivettp: we seem to be at cross purposes. My comment was on property hasReferenceDateAndTime and the fact that its definition "specifies a reference date...for a given value" was circular. The definition should be something like "the time when the value actually occurred"

IND/EconomicIndicators/EconomicIndicators.rdf
rivettp: You didn't address my 2nd point which is that the direction should be that the value should reference the Index; we should not encourage people to try to link from the Index to reference all its values over time.

ElisaKendall: Let's do this in another issue - I think more work may be involved to clean some of this up.

FND/Utilities/Analytics.rdf
rivettp: Still think isValueOf (good name change) should have cardinality of 1.

IND/EconomicIndicators/EconomicIndicators.rdf
```diff
@@ -388,28 +386,26 @@
...
+				<owl:onProperty rdf:resource="&fibo-fnd-utl-alx;hasArgument"/>
 				<owl:someValuesFrom rdf:resource="&fibo-ind-ei-ei;CivilianNonInstitutionalPopulation"/>
 			</owl:Restriction>
 		</rdfs:subClassOf>
 		<rdfs:subClassOf>
 			<owl:Restriction>
-				<owl:onProperty rdf:resource="&fibo-fnd-utl-alx;hasArgument"/>
-				<owl:someValuesFrom rdf:resource="&fibo-fbc-pas-fpas;Basket"/>
+				<owl:onProperty rdf:resource="&fibo-ind-ei-ei;hasIndicatorValue"/>
+				<owl:someValuesFrom rdf:resource="&fibo-fnd-utl-alx;NumericIndexValue"/>
```
rivettp, line 408: As discussed I think it not a good idea to, not only encourage but, require triples to be established in this direction (from Indicator to potentially hundreds of values).

Issue #1015, opened 2020-04-26
Augment the equities concepts with their CFI classification


ElisaKendall:
The concept of the a CFI classifier and classification scheme should be added under SEC/Securities, with proper named individuals of type CFIClassifier added as value restrictions on the definitions of the appropriate instruments.

Most of the instruments listed in the CFI standard under equities are already in the released ontologies. In some cases, instrument descriptions provided in the ISO standard should be added as explanatory notes, and additional restrictions may be needed to further refine some of the concepts.

ElisaKendall:
Reopening to address more changes per the SEC FCT

PR #1016, opened 2020-04-26
SEC-73 - Augment the equities concepts with their CFI classification

ElisaKendall:

## Description

This initial PR reflects work-in-progress by the Securities FCT to augment the securities and equities ontologies to facilitate incorporation of the CFI codes for equities.  They represent a small subset of the CFI standard, but are important for several organizations participating in the FCT.

This initial pull request includes the additional concepts identified by the FCT to enable representation of the codes.  A subsequent pull request will include examples of the codes themselves classifying specific example instruments, to show FIBO users how to encode them for various applications.

Fixes: #1015  / SEC-73

## Checklist:

### DIFF
--- a/SEC/Equities/EquityInstruments.rdf
+++ b/SEC/Equities/EquityInstruments.rdf
@@ -133,6 +133,19 @@
 		<fibo-fnd-utl-av:synonym xml:lang="en-GB">ordinary share</fibo-fnd-utl-av:synonym>
 	</owl:Class>
 	
+	<owl:Class rdf:about="&fibo-sec-eq-eq;ConvertibleCommonShare">
+		<rdfs:subClassOf rdf:resource="&fibo-sec-eq-eq;CommonShare"/>
+		<rdfs:subClassOf rdf:resource="&fibo-sec-sec-iss;ConvertibleSecurity"/>
+		<rdfs:subClassOf>
+			<owl:Restriction>
+				<owl:onProperty rdf:resource="&fibo-fnd-agr-ctr;hasContractualElement"/>
+				<owl:someValuesFrom rdf:resource="&fibo-sec-eq-eq;EquityConversionTerms"/>
+			</owl:Restriction>
+		</rdfs:subClassOf>
+		<rdfs:label>convertible common share</rdfs:label>
+		<skos:definition>common share that is convertible into another security</skos:definition>
+	</owl:Class>
+	
 	<owl:Class rdf:about="&fibo-sec-eq-eq;ConvertiblePreferredShare">
 		<rdfs:subClassOf rdf:resource="&fibo-sec-eq-eq;PreferredShare"/>
 		<rdfs:subClassOf rdf:resource="&fibo-sec-sec-iss;ConvertibleSecurity"/>
@@ -143,7 +156,7 @@
 			</owl:Restriction>
 		</rdfs:subClassOf>
 		<rdfs:label>convertible preferred share</rdfs:label>
-		<skos:definition>a preferred share that is convertible into another security</skos:definition>
+		<skos:definition>preferred share that is convertible into another security</skos:definition>
 	</owl:Class>
 	
 	<owl:Class rdf:about="&fibo-sec-eq-eq;CumulativePreferredShare">
@@ -217,6 +230,12 @@
 		<rdfs:subClassOf rdf:resource="&fibo-sec-sec-sch;ParametricSchedule"/>
 		<rdfs:label>dividend or share coupon schedule</rdfs:label>
 		<skos:definition>payment schedule indicating the dates on which dividends or share coupon payments are due to be paid</skos:definition>
+	</owl:Class>
+	
+	<owl:Class rdf:about="&fibo-sec-eq-eq;EnhancedVoting">
+		<rdfs:subClassOf rdf:resource="&fibo-sec-eq-eq;VotingRight"/>
+		<rdfs:label>enhanced voting</rdfs:label>
+		<skos:definition>shareholder is entitled to more than one vote per share</skos:definition>
 	</owl:Class>
 	
 	<owl:Class rdf:about="&fibo-sec-eq-eq;EquityConversionTerms">
@@ -386,6 +405,18 @@
 		<skos:definition>preferred share that is not a participating preferred share</skos:definition>
 	</owl:Class>
 	
+	<owl:Class rdf:about="&fibo-sec-eq-eq;NonVoting">
+		<rdfs:subClassOf rdf:resource="&fibo-sec-eq-eq;VotingRight"/>
+		<rdfs:label>non-voting</rdfs:label>
+		<owl:equivalentClass>
+			<owl:Restriction>
+				<owl:onProperty rdf:resource="&fibo-sec-eq-eq;confersNumberOfVotesPerShare"/>
+				<owl:hasValue rdf:datatype="&xsd;decimal">0</owl:hasValue>
+			</owl:Restriction>
+		</owl:equivalentClass>
+		<skos:definition>share has no voting rights</skos:definition>
+	</owl:Class>
+	
 	<owl:Class rdf:about="&fibo-sec-eq-eq;OrdinaryDividend">
 		<rdfs:subClassOf rdf:resource="&fibo-sec-eq-eq;Dividend"/>
 		<rdfs:label>ordinary dividend</rdfs:label>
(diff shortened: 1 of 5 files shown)


SEC/Equities/EquityInstruments.rdf
```diff
@@ -108,17 +116,47 @@
...
 		<fibo-fnd-utl-av:hasMaturityLevel rdf:resource="&fibo-fnd-utl-av;Release"/>
 	</owl:Ontology>
 	
+	<owl:Class rdf:about="&fibo-sec-eq-eq;CommonNonVotingUnrestrictedFullyPaidRegisteredShare">
+		<rdfs:subClassOf rdf:resource="&fibo-sec-eq-eq;CommonShare"/>
+		<rdfs:subClassOf rdf:resource="&fibo-sec-eq-eq;UnrestrictedShare"/>
+		<rdfs:subClassOf rdf:resource="&fibo-sec-sec-lst;RegisteredSecurity"/>
+		<rdfs:subClassOf>
+			<owl:Restriction>
+				<owl:onProperty rdf:resource="&lcc-cr;isClassifiedBy"/>
```
rivettp, line 135: I think this is the wrong property to use: there should be soemthing more specific.

SEC/Equities/EquityInstruments.rdf
```diff
@@ -108,17 +116,47 @@
...
+				<owl:onClass rdf:resource="&fibo-sec-eq-eq;FullyPaidShareStatus"/>
+				<owl:qualifiedCardinality rdf:datatype="&xsd;nonNegativeInteger">1</owl:qualifiedCardinality>
+			</owl:Restriction>
+		</rdfs:subClassOf>
+		<rdfs:subClassOf>
+			<owl:Restriction>
+				<owl:onProperty rdf:resource="&fibo-fnd-rel-rel;confers"/>
```
rivettp, line 142: Again I don't think this is the right property, especially with respect to NonVoting where literally nothing is conferred - the restriction should be a cardinality 0 restriction onClass VotingRight.

SEC/Equities/EquityInstruments.rdf
```diff
@@ -108,17 +116,47 @@
...
+			</owl:Restriction>
+		</rdfs:subClassOf>
+		<rdfs:subClassOf>
+			<owl:Restriction>
+				<owl:onProperty rdf:resource="&fibo-sec-sec-iss;isIssuedInForm"/>
+				<owl:onClass rdf:resource="&fibo-sec-sec-lst;RegisteredSecurityForm"/>
+				<owl:qualifiedCardinality rdf:datatype="&xsd;nonNegativeInteger">1</owl:qualifiedCardinality>
+			</owl:Restriction>
+		</rdfs:subClassOf>
+		<rdfs:label xml:lang="en-US">common, non-voting, unrestricted, fully-paid, registered share</rdfs:label>
```
rivettp, line 154: We should include the 6 character CFI code as an abbreviation or altLabel, or is this represented another way? SHould have an explanatoryNote?

SEC/Equities/EquityInstruments.rdf
```diff
@@ -530,10 +658,29 @@
 				<owl:someValuesFrom rdf:resource="&fibo-fnd-acc-aeq;ShareholdersEquity"/>
 			</owl:Restriction>
 		</rdfs:subClassOf>
+		<rdfs:subClassOf>
+			<owl:Restriction>
+				<owl:onProperty rdf:resource="&lcc-cr;isClassifiedBy"/>
```
rivettp, line 663: again I'm not keen on use of isClassifiedBy

SEC/Securities/SecuritiesClassification.rdf
```diff
@@ -85,6 +89,43 @@
...
+	<owl:Class rdf:about="&fibo-sec-sec-cls;ClassificationOfFinancialInstrumentsCodeScheme">
+		<rdfs:subClassOf rdf:resource="&fibo-sec-sec-cls;FinancialInstrumentClassificationScheme"/>
+		<rdfs:subClassOf rdf:resource="&lcc-lr;CodeSet"/>
+		<rdfs:subClassOf>
+			<owl:Restriction>
+				<owl:onProperty rdf:resource="&fibo-fnd-rel-rel;defines"/>
+				<owl:someValuesFrom rdf:resource="&fibo-sec-sec-cls;FinancialInstrumentClassificationCode"/>
+			</owl:Restriction>
+		</rdfs:subClassOf>
+		<rdfs:label>classification of financial instruments code scheme</rdfs:label>
```
rivettp, line 101: Include CFI Code as an abbreviation

SEC/Securities/SecuritiesIssuance.rdf
```diff
@@ -547,6 +615,14 @@
 		<skos:definition>indicates a period of time in which investors can commit to purchase shares (or units) to be issued</skos:definition>
 	</owl:ObjectProperty>
 	
+	<owl:ObjectProperty rdf:about="&fibo-sec-sec-iss;isIssuedInForm">
+		<rdfs:subPropertyOf rdf:resource="&fibo-fnd-agr-ctr;isEvidencedBy"/>
```
rivettp, line 619: I don't get use of this super property, though isEvidencedBy is very poorly defined as "is attested by", there is no way the SecurityForm is evidence for anything

SEC/Securities/SecuritiesIssuance.rdf
```diff
@@ -547,6 +615,14 @@
...
+		<rdfs:label>is issued in form</rdfs:label>
+		<rdfs:domain rdf:resource="&fibo-fbc-fi-fi;Security"/>
+		<rdfs:range rdf:resource="&fibo-sec-sec-iss;SecurityForm"/>
```
rivettp, line 622: Major problem - this means that BearerForm, RegisteredForm should be Individuals not subClasses or you get punning.

SEC/Securities/SecuritiesIssuance.rdf
```diff
@@ -547,6 +615,14 @@
...
+		<skos:definition>indicates the form in which the security is issued, typically in registered form</skos:definition>
+	</owl:ObjectProperty>
+	
 	<owl:DatatypeProperty rdf:about="&fibo-sec-sec-iss;isOverAlloted">
```
rivettp, line 626: typo should be isOverAllotted (double t)

SEC/Securities/SecuritiesListings.rdf
```diff
@@ -235,11 +232,39 @@
 				<owl:someValuesFrom rdf:resource="&fibo-fnd-dt-fd;CombinedDateTime"/>
 			</owl:Restriction>
 		</rdfs:subClassOf>
+		<rdfs:subClassOf>
+			<owl:Restriction>
+				<owl:onProperty rdf:resource="&fibo-sec-sec-iss;isIssuedInForm"/>
+				<owl:someValuesFrom rdf:resource="&fibo-sec-sec-lst;RegisteredSecurityForm"/>
```
rivettp, line 238: how could you have an instance of RegisteredSecurityForm - it should be an individual ( required to avoid punning - see earlier comment)

SEC/Equities/EquityInstruments.rdf
ElisaKendall: I'm thinking the codes need to be in a separate ontology, not on the class in this ontology.  We have not decided to have an exhaustive set of classes for the CFI standard - doing so would take weeks to get right, and would add a lot of unneeded content.  However, providing the codes using the patterns we have done previously, stating the combination of characteristics they should have in text, and mapping those that we don't have subclasses for to the higher level class seems like a reasonable approach.

SEC/Securities/SecuritiesIssuance.rdf
ElisaKendall: ok

SEC/Equities/EquityInstruments.rdf
ElisaKendall: Using cardinality here would give you the wrong semantics.  In fact, what we want to say here is that this class of shares is non-voting, in other words, confers no voting rights to its shareholders.  Thus, the restriction says that the kind of voting rights this share confers is exactly non-voting.  Then we say that non-voting voting rights means that the shareholder gets precisely zero votes.

You could argue that we should make these classes of voting rights individuals, but then we would have to do something clever with the two cases in which we cannot give an exact number of voting rights, which would apply to Alphabet Class B shares - which confer 10 votes per share.

SEC/Equities/EquityInstruments.rdf
ElisaKendall: Ok - I've created a subproperty of SharePaymentStatus called hasSharePaymentStatus, and have made the 3 options individuals.

SEC/Equities/EquityInstruments.rdf
ElisaKendall: fixed - see above

SEC/Securities/SecuritiesIssuance.rdf
ElisaKendall: The contract document, either virtual or physical, such as a bearer bond, provides evidence for ownership.  How that evidence is manifested, i.e., the form it takes, is what this property is intended to be used for.  So that's why I selected isEvidencedBy, but I understand, I think, what you are saying here.  I couldn't use takesForm from FND, which has a very specific meaning, so I've replaced it with isRealizedBy.

SEC/Securities/SecuritiesIssuance.rdf
ElisaKendall: I've addressed this through a series of changes to how these are used - they were already individuals, but the restrictions didn't use them consistently.  Took a couple of passes to get this right.

SEC/Securities/SecuritiesListings.rdf
ElisaKendall: fixed - see above

SEC/Equities/EquityInstruments.rdf
rivettp: why is 10 votes not exact?

SEC/Equities/EquityInstruments.rdf
rivettp: Can you give an example of one code so I can get the idea?

SEC/Equities/EquityInstruments.rdf
ElisaKendall: If you encoded it that way, you would blow up the search space and no reasoner or rule engine would be able to handle it.  The right way to do that would be to make an individual of type EnhancedVoting and then use the fibo-sec-eq-eq;confersNumberOfVotesPerShare property with a value of 10 on the instance.  The individual could be called something like 'enhanced voting with 10 votes per share'.  Then the individual for the shares, in the case of alphabet class b shares, would likely be a member of a convertible preferred share with the value of 'enhanced voting with 10 votes per share' for the confers voting rights.

SEC/Equities/EquityInstruments.rdf
ElisaKendall: Yes, next week.  I'll work on the codes ontology when the conference is over.

Issue #1041, opened 2020-06-13
Partnerships are not well specified and partnership with general partners is overly constrained


ElisaKendall:
Many partnerships have at least one general partner, but a partnership with general partners doesn't necessarily have to have at least two general partners. Also, many partnerships are registered but not incorporated (only corporations are 'incorporated'), with partners that may or may not be jointly and severally liable, and they are all established via a partnership agreement, so the classes called non-incorporated partnership and partnership incorporated through agreement are non-sensical and should be eliminated.

A partnership with general and limited partners, such as an LLLP, may have one managing / general partner and multiple limited partners, so the restriction on the number of required general partners on its parent class is a problem, and we also need the subclass of a limited liability limited partnership.

'partnership with general partners with limited liability' should be renamed limited liability partnership. 'partnership with only limited partners' should be renamed 'limited partnership'.

PR #1042, opened 2020-06-13
BE-216 - Partnerships are not well specified and partnership with general partners is overly constrained

ElisaKendall:

## Description

Streamlined the partnerships ontology, eliminating categories of partnership that were defined exhaustively, but for which there are no known examples world-wide, loosening / refining restrictions and definitions, added new concepts and related definitions to correspond to the kinds of partnerships that actually do exist together with the roles that partners play in them, reworked definitions to be ISO 704 compliant and more representative of the concepts they apply to.

Fixes: #1041 / BE-216

## Checklist:

### DIFF
--- a/BE/Partnerships/Partnerships.rdf
+++ b/BE/Partnerships/Partnerships.rdf
@@ -274,6 +274,7 @@
 	<owl:ObjectProperty rdf:about="&fibo-be-ptr-ptr;hasGeneralPartner">
 		<rdfs:subPropertyOf rdf:resource="&fibo-fnd-pty-pty;isAffectedBy"/>
 		<rdfs:label>has general partner</rdfs:label>
+		<rdfs:domain rdf:resource="&fibo-be-ptr-ptr;Partnership"/>
 		<rdfs:range rdf:resource="&fibo-be-ptr-ptr;GeneralPartner"/>
 		<skos:definition>indicates an actor that has some measure of control over the partnership</skos:definition>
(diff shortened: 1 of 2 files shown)


dallemang:
Is there a reason why you eliminated Corporate Limited Partner? This situation is possible (and, as per the lawyer at WF where we got the idea, somewhat common), where a corporation enters into a partnership as a limited liability member.   This situation is not outlawed by the proposed change, but we no longer have a particular class for this situation, so we can't talk about, for example, any regulations that might apply differently to Natural Persons vs. Legal Entities (Corporations) when they are limited partners.    I'm not sure what the use case was for including them (@dsnewman might remember).

ElisaKendall:
> Is there a reason why you eliminated Corporate Limited Partner? This situation is possible (and, as per the lawyer at WF where we got the idea, somewhat common), where a corporation enters into a partnership as a limited liability member. This situation is not outlawed by the proposed change, but we no longer have a particular class for this situation, so we can't talk about, for example, any regulations that might apply differently to Natural Persons vs. Legal Entities (Corporations) when they are limited partners. I'm not sure what the use case was for including them (@dsnewman might remember).

@dallemang - limited partners can be either people or corporations, so if we need to keep this, which seemed like overkill, then we would need to add the parallel case where partners can be people.  The ontology distinguished between natural person partnerships (not partners) and corporate limited partners previously, but we can talk about which is which via restrictions on their identity, so it's redundant.

Issue #1044, opened 2020-06-23
The definition of a signatory is disconnected from legally delegated authority and responsible party


ElisaKendall:
The definition of signatory is disconnected from concepts it should depend on including legally delegated authority and responsible party in Executives. There are two redundant properties linking an organization or legal entity to a signatory and neither allows a person to delegate a signatory via power of attorney. They should be merged and redefined appropriately to cover the power of attorney case as well as a corporate designated signatory case. Doing so requires moving these concepts/properties from LegalPersons to Executives in BE.

Further, the concept of having signing authority is temporal, and should be integrated with a legal authorization situation that has a time frame. For the case of a power of attorney, there may or may not be an end date or trigger for an end date, and similarly for a signer on an account, so the time frame should be optional, but likely will have a start date.

PR #1045, opened 2020-06-24
BE-217 - The definition of a signatory is disconnected from legally delegated authority and responsible party

ElisaKendall:

## Description

Moved signatory related concepts to the executives ontology; in the context of the executives ontology: integrated the concept of authorization, authorizing/authorized parties, and related properties to enable better semantics of signatories and executives, simplified the ontology to eliminate some of the concepts around corporate officer duty and related obligations, revised definitions to be ISO 704 compliant

Fixes: #1044 / BE-217

## Checklist:

### DIFF
--- a/BE/OwnershipAndControl/Executives.rdf
+++ b/BE/OwnershipAndControl/Executives.rdf
@@ -82,12 +82,13 @@
 		<owl:imports rdf:resource="https://spec.edmcouncil.org/fibo/ontology/FND/Relations/Relations/"/>
 		<owl:imports rdf:resource="https://spec.edmcouncil.org/fibo/ontology/FND/Utilities/AnnotationVocabulary/"/>
 		<owl:imports rdf:resource="https://www.omg.org/spec/LCC/Languages/LanguageRepresentation/"/>
-		<owl:versionIRI rdf:resource="https://spec.edmcouncil.org/fibo/ontology/BE/20200201/OwnershipAndControl/Executives/"/>
+		<owl:versionIRI rdf:resource="https://spec.edmcouncil.org/fibo/ontology/BE/20200601/OwnershipAndControl/Executives/"/>
 		<skos:changeNote>The https://spec.edmcouncil.org/fibo/ontology/BE/20131101/OwnershipAndControl/Executives.rdf version of this ontology was modified per the issue resolutions identified in the FIBO BE 1.0 FTF report.</skos:changeNote>
 		<skos:changeNote>The https://spec.edmcouncil.org/fibo/ontology/BE/20160801/OwnershipAndControl/Executives.rdf version of this ontology was modified per the FIBO 2.0 RFC to fix reasoning issues.</skos:changeNote>
 		<skos:changeNote>The https://spec.edmcouncil.org/fibo/ontology/BE/20180801/OwnershipAndControl/Executives.rdf version of this ontology was modified to clarify the definition of responsible party.</skos:changeNote>
 		<skos:changeNote>The https://spec.edmcouncil.org/fibo/ontology/BE/20181101/OwnershipAndControl/Executives.rdf version of this ontology was modified to adjust the hierarchy to better support regulatory requirements.</skos:changeNote>
 		<skos:changeNote>The https://spec.edmcouncil.org/fibo/ontology/BE/20190501/OwnershipAndControl/Executives.rdf version of this ontology was modified to eliminate duplication with concepts in LCC.</skos:changeNote>
+		<skos:changeNote>The https://spec.edmcouncil.org/fibo/ontology/BE/20200201/OwnershipAndControl/Executives.rdf version of this ontology was modified to integrate concepts related to authorization, including the concept of a signatory (moved from legal persons) to improve semantics; simplify the ontology, and normalize definitions to be ISO 704 compliant.</skos:changeNote>
 		<fibo-fnd-utl-av:hasMaturityLevel rdf:resource="&fibo-fnd-utl-av;Release"/>
 	</owl:Ontology>
 	
@@ -95,11 +96,10 @@
 		<rdfs:subClassOf>
 			<owl:Restriction>
 				<owl:onProperty rdf:resource="&fibo-fnd-rel-rel;governs"/>
-				<owl:onClass rdf:resource="&fibo-be-oac-exec;CompanyBylaw"/>
-				<owl:qualifiedCardinality rdf:datatype="&xsd;nonNegativeInteger">1</owl:qualifiedCardinality>
+				<owl:someValuesFrom rdf:resource="&fibo-be-oac-exec;Bylaws"/>
 			</owl:Restriction>
 		</rdfs:subClassOf>
-		<fibo-fnd-utl-av:explanatoryNote>extends the definition of an instrument of incorporation to constrain corporate bylaws</fibo-fnd-utl-av:explanatoryNote>
+		<fibo-fnd-utl-av:explanatoryNote>An instrument of incorporation also constrains corporate bylaws.</fibo-fnd-utl-av:explanatoryNote>
 	</owl:Class>
 	
 	<owl:Class rdf:about="&fibo-be-oac-cpty;DeFactoControllingInterestParty">
@@ -110,7 +110,7 @@
 				<owl:minQualifiedCardinality rdf:datatype="&xsd;nonNegativeInteger">0</owl:minQualifiedCardinality>
 			</owl:Restriction>
 		</rdfs:subClassOf>
-		<fibo-fnd-utl-av:explanatoryNote>a control owner (i.e., control person, per SEC regulations) may have some means or right that allows them to exercise control over board composition, other than through proxy assignment or vote.  Not all control persons have this facility, as it is not inherent to having a significant (for example, 20 percent or more) ownership stake.</fibo-fnd-utl-av:explanatoryNote>
+		<fibo-fnd-utl-av:explanatoryNote>A control owner (i.e., control person, per SEC regulations) may have some means or right that allows them to exercise control over board composition, other than through proxy assignment or vote.  Not all control persons have this facility, as it is not inherent to having a significant (for example, 20 percent or more) ownership stake.</fibo-fnd-utl-av:explanatoryNote>
 	</owl:Class>
 	
 	<owl:Class rdf:about="&fibo-be-oac-cctl;VotingShareholder">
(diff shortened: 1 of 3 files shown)


BE/OwnershipAndControl/Executives.rdf
```diff
@@ -177,18 +233,13 @@
...
 		<rdfs:subClassOf rdf:resource="&fibo-fnd-pty-pty;OrganizationMember"/>
-		<rdfs:subClassOf>
-			<owl:Restriction>
-				<owl:onProperty rdf:resource="&fibo-fnd-rel-rel;hasIdentity"/>
-				<owl:allValuesFrom rdf:resource="&fibo-be-le-lp;LegalPerson"/>
-			</owl:Restriction>
-		</rdfs:subClassOf>
 		<rdfs:subClassOf>
 			<owl:Restriction>
 				<owl:onProperty rdf:resource="&fibo-fnd-pty-rl;isPlayedBy"/>
```
rivettp, line 245: This restriction on isAPartyTo is a bit pointless, since it does not constrain to the Agreement of the same company. In general I think it's safe to assume that any board member is party to such an agreement - it's not FIBO's job to track each individual who actually signed!
Likewise the restriction on isPlayedBy has a lot of complexity and still does not constrain membership to the board of the same company.
In summary I don't think we need wither - just a reified relationship (giving their term etc. ) from Board to LegalPerson

BE/OwnershipAndControl/Executives.rdf
```diff
@@ -256,171 +307,93 @@
...
-		<fibo-fnd-utl-av:adaptedFrom rdf:datatype="&xsd;anyURI">http://www.businessdictionary.com/definition/bylaw.html</fibo-fnd-utl-av:adaptedFrom>
+		<rdfs:label>bylaws</rdfs:label>
+		<skos:definition>written rules for conduct of a corporation, association, partnership or any organization</skos:definition>
+		<fibo-fnd-utl-av:explanatoryNote>Corporate bylaws are typically called Articles of Association in the United Kingdom, or Articles of Incorporation in the United States. In government usage, the term bylaws is used to describe laws, orders, regulations, rules, etc., made by a ministry, local authority (a municipality, for example), or public corporation, in accordance with the powers conferred by or delegated under a statue (called the parent act).</fibo-fnd-utl-av:explanatoryNote>
```
rivettp, line 318: I think Articles and Bylaws are different: indeed we have a Property (mentioned earlier) to link them (though I don't think the property is useful). See https://www.upcounsel.com/bylaws-vs-articles-of-incorporation#:~:text=The%20Articles%20of%20Incorporation%20include,procedures%20for%20managing%20the%20corporation.

Maybe have as synonym "membership agreement"?

BE/OwnershipAndControl/Executives.rdf
```diff
@@ -256,171 +307,93 @@
...
 	</owl:Class>
 	
-	<owl:Class rdf:about="&fibo-be-oac-exec;Bylaw">
+	<owl:Class rdf:about="&fibo-be-oac-exec;Bylaws">
```
rivettp, line 314: Why plural? What if there are 20 Bylaws - will they all be grouped into a single resource?

BE/OwnershipAndControl/Executives.rdf
```diff
@@ -256,171 +307,93 @@
...
 	</owl:Class>
 	
 	<owl:Class rdf:about="&fibo-be-oac-exec;ChiefFinancialOfficer">
 		<rdfs:subClassOf rdf:resource="&fibo-be-oac-exec;CorporateOfficer"/>
 		<rdfs:subClassOf rdf:resource="&fibo-be-oac-exec;Executive"/>
 		<rdfs:label>chief financial officer</rdfs:label>
-		<skos:definition>Senior-most executive responsible for financial control and planning of a firm or project. He or she is in charge of all accounting functions including (1) credit control, (2) preparing budgets and financial statements, (3) coordinating financing and fund raising, (4) monitoring expenditure and liquidity, (5) managing investment and taxation issues, (6) reporting financial performance to the board, and (7) providing timely financial data to the CEO. Also called chief finance officer, comptroller, controller, or finance controller.</skos:definition>
+		<skos:definition>senior-most corporate officer responsible for financial control and planning of organization or project</skos:definition>
```
rivettp, line 347: "organization or project" at the end does not read right

BE/OwnershipAndControl/Executives.rdf
```diff
@@ -256,171 +307,93 @@
...
 	</owl:Class>
 	
 	<owl:Class rdf:about="&fibo-be-oac-exec;CompanyLaw">
 		<rdfs:subClassOf rdf:resource="&fibo-fnd-law-jur;StatuteLaw"/>
 		<rdfs:label>company law</rdfs:label>
-		<skos:definition>Legislation under which the formation, registration or incorporation, governance, and dissolution of a firm is administered and controlled.</skos:definition>
-		<fibo-fnd-utl-av:definitionOrigin rdf:datatype="&xsd;anyURI">http://www.businessdictionary.com/definition/company-law.html</fibo-fnd-utl-av:definitionOrigin>
+		<skos:definition>legislation under which the formation, registration or incorporation, governance, and dissolution of a firm is administered and controlled</skos:definition>
```
rivettp, line 356: Would an instance represent one specific law, or the corpus of laws (for a specific jurisdiction?).
To help answer, What is the use case for this? I cannot see any purpose for the former, only specific laws. But don't think we have sufficient to support meaningful use cases.

BE/OwnershipAndControl/Executives.rdf
```diff
@@ -256,171 +307,93 @@
...
+		<skos:definition>corporate officer appointed by the directors of an organization, responsible for ensuring compliance with legal obligations related to corporate governance</skos:definition>
+		<fibo-fnd-utl-av:adaptedFrom rdf:datatype="&xsd;anyURI">http://www.businessdictionary.com/definition/company-secretary.html</fibo-fnd-utl-av:adaptedFrom>
+		<fibo-fnd-utl-av:explanatoryNote>His or her formal duties include (1) calling meetings, (2) recording minutes of the meetings, (3) keeping statutory record books, (4) proper payment of dividend and interest payments, and (5) proper drafting and execution of agreements, contracts, and resolutions. A company secretary is not automatically an employee of the firm and, if employed with executive responsibilities, not be its director shareholder. If a firm has only two directors, one may act as its secretary; but a sole director may not. A firm (such as of accountants) may not act as a company secretary for any firm. Called corporate secretary in the US.</fibo-fnd-utl-av:explanatoryNote>
```
rivettp, line 366: what's "director shareholder"?

BE/OwnershipAndControl/Executives.rdf
```diff
@@ -256,171 +307,93 @@
...
-	<owl:Class rdf:about="&fibo-be-oac-exec;CorporateOfficerDuty">
-		<rdfs:subClassOf rdf:resource="&fibo-be-oac-exec;CompanyStatutoryObligation"/>
-		<rdfs:label>corporate officer duty</rdfs:label>
-		<skos:definition>some duty incumbent upon some officer of a company</skos:definition>
+		<skos:definition>high-level management executive of a corporation or other organization, hired by the board of directors or the business owner(s), charged with certain operational responsibilities</skos:definition>
+		<skos:example>Corporate officers may include a Chief Executive Officer (CEO), Chief Financial Officer (CFO), president, vice president(s), and in some cases a Chief Operating Officer (COO), Chief Compliance Officer (CCO), or other executive responsible for a critical function in the organization.</skos:example>
 	</owl:Class>
 	
 	<owl:Class rdf:about="&fibo-be-oac-exec;Executive">
 		<rdfs:subClassOf rdf:resource="&fibo-be-oac-exec;LegallyDelegatedAuthority"/>
```
rivettp, line 379: Surely they HAVE authority not ARE authority.
Is there no realtionship with Corporate Officer - are they orthogonal?

BE/OwnershipAndControl/Executives.rdf
```diff
@@ -256,171 +307,93 @@
...
+		<skos:definition>person appointed and given the responsibility to manage the affairs of an organization and the authority to make decisions within specified role-specific boundaries</skos:definition>
+		<fibo-fnd-utl-av:adaptedFrom rdf:datatype="&xsd;anyURI">http://www.businessdictionary.com/definition/executive.html</fibo-fnd-utl-av:adaptedFrom>
 	</owl:Class>
 	
 	<owl:Class rdf:about="&fibo-be-oac-exec;ExecutiveBoardMember">
 		<rdfs:subClassOf rdf:resource="&fibo-be-oac-exec;BoardMember"/>
 		<rdfs:subClassOf rdf:resource="&fibo-be-oac-exec;Executive"/>
 		<rdfs:label>executive board member</rdfs:label>
-		<skos:definition>An individual performing the role of a board member of a company who also has executive responsibilities in the company.</skos:definition>
+		<skos:definition>executive that is also a member of the board of directors of the organization they manage</skos:definition>
```
rivettp, line 396: circular

BE/OwnershipAndControl/Executives.rdf
```diff
@@ -463,10 +436,45 @@
...
+					</owl:Restriction>
+				</owl:someValuesFrom>
+			</owl:Restriction>
+		</rdfs:subClassOf>
+		<rdfs:label>signatory</rdfs:label>
+		<skos:definition>responsible party authorized to sign agreements on behalf of themselves, another person, or an organization</skos:definition>
 	</owl:Class>
 	
+	<owl:ObjectProperty rdf:about="&fibo-be-oac-exec;authorizes">
+		<rdfs:subPropertyOf rdf:resource="&fibo-fnd-pty-pty;actsOn"/>
```
rivettp, line 461: does not seem right

BE/OwnershipAndControl/Executives.rdf
```diff
@@ -475,17 +483,40 @@
...
+	</owl:ObjectProperty>
+	
 	<owl:ObjectProperty rdf:about="&fibo-be-oac-exec;elects">
 		<rdfs:label>elects</rdfs:label>
-		<rdfs:domain rdf:resource="&fibo-fnd-pty-pty;PartyInRole"/>
-		<rdfs:range rdf:resource="&fibo-fnd-pty-pty;PartyInRole"/>
+		<rdfs:domain rdf:resource="&fibo-be-oac-exec;AuthorizingParty"/>
+		<rdfs:range rdf:resource="&fibo-be-oac-exec;AuthorizedParty"/>
 		<skos:definition>chooses someone to hold office or some other position by voting</skos:definition>
 		<skos:editorialNote>In the case of an election for a member of the board of directors, the bylaws state the manner in which that process is effected. The candidate members may be recommended by the board or other proxy and are then elected by the shareholder. A similar process may be conducted to elect outside auditors.</skos:editorialNote>
```
rivettp, line 498: shareholder should be plural. 
For a public company with millions of shareholders would each be linked to a BoardMember using this property, assuming they voted? I don't FIBO is intended for use to tally votes!
As per previous comment need to distinguish appoints from elects.
Overall I don't think this property is needed at all beyond a general need to track where someone has control and appoints a director.

BE/OwnershipAndControl/Executives.rdf
```diff
@@ -520,6 +551,14 @@
...
 	
+	<owl:ObjectProperty rdf:about="&fibo-be-oac-exec;hasSigningAuthorityFor">
+		<rdfs:subPropertyOf rdf:resource="&fibo-be-oac-exec;isAuthorizedBy"/>
+		<rdfs:label>has signing authority for</rdfs:label>
+		<rdfs:domain rdf:resource="&fibo-be-oac-exec;Signatory"/>
+		<owl:inverseOf rdf:resource="&fibo-be-oac-exec;designatesSignatory"/>
+		<skos:definition>indicates the party for which a signatory has the ability to sign agreements, access accounts and perform related tasks</skos:definition>
+	</owl:ObjectProperty>
+	
 	<owl:ObjectProperty rdf:about="&fibo-be-oac-exec;hasVestedInIt">
```
rivettp, line 562: poorly worded and defined, is it needed?

BE/OwnershipAndControl/Executives.rdf
```diff
@@ -127,25 +127,81 @@
...
+		<skos:definition>party qualified and authorized to review and verify the accuracy of financial records and ensure that companies comply with tax laws</skos:definition>
+		<fibo-fnd-utl-av:explanatoryNote>An auditor may be an internal auditor - an individual whose primary job function is to audit his or her own company, or an external auditor - an individual from outside the company, who typically is employed by an auditing firm who handles many different clients.</fibo-fnd-utl-av:explanatoryNote>
+	</owl:Class>
+	
+	<owl:Class rdf:about="&fibo-be-oac-exec;Authorization">
+		<rdfs:subClassOf rdf:resource="&fibo-fnd-pty-pty;Situation"/>
+		<rdfs:subClassOf>
+			<owl:Restriction>
+				<owl:onProperty rdf:resource="&fibo-be-oac-exec;hasAuthorizedParty"/>
+				<owl:someValuesFrom rdf:resource="&fibo-be-oac-exec;AuthorizedParty"/>
```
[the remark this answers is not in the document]
dallemang: Something that we have found in doing some mapping for Wells Fargo is the advantage of using the same (or similar) pattern for lots of relations.   While it is possible to mix 3-class and 5-class patterns, it is fussy.  For example, WF represents all the relations it knows (ownership, beneficial ownership, signatory power, control, and a dozen others) as being between one customer and another, where the customer is clearly a relative thing (it has its own number; they don't use the LEI or the DUNS for this).  So we manage a network of relative things (that's what they are already doing).  If we represent some of the relations as 3-class, we don't have the relative things to link it, so we would have to combine independent things with the relative things.  This is probably more accurate (after all, this entity owns that entity, regardless of whether either or both are our customers), but conceptually difficult - when am I talking about the customer and when am I talking about the entity?  If we use the same 5-class pattern for all of them, we at least have a place for the corresponding things to be.   So, while it isn't conceptually necessary, it makes it easier to compare and combine data about different kinds of relationships.

BE/OwnershipAndControl/Executives.rdf
```diff
@@ -82,24 +82,24 @@
...
 		<rdfs:subClassOf>
 			<owl:Restriction>
 				<owl:onProperty rdf:resource="&fibo-fnd-rel-rel;governs"/>
-				<owl:onClass rdf:resource="&fibo-be-oac-exec;CompanyBylaw"/>
-				<owl:qualifiedCardinality rdf:datatype="&xsd;nonNegativeInteger">1</owl:qualifiedCardinality>
+				<owl:someValuesFrom rdf:resource="&fibo-be-oac-exec;Bylaws"/>
 			</owl:Restriction>
 		</rdfs:subClassOf>
-		<fibo-fnd-utl-av:explanatoryNote>extends the definition of an instrument of incorporation to constrain corporate bylaws</fibo-fnd-utl-av:explanatoryNote>
+		<fibo-fnd-utl-av:explanatoryNote>An instrument of incorporation also constrains corporate bylaws.</fibo-fnd-utl-av:explanatoryNote>
```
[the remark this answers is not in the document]
ElisaKendall: I removed the restriction.

BE/OwnershipAndControl/Executives.rdf
```diff
@@ -110,7 +110,7 @@
 				<owl:minQualifiedCardinality rdf:datatype="&xsd;nonNegativeInteger">0</owl:minQualifiedCardinality>
 			</owl:Restriction>
 		</rdfs:subClassOf>
-		<fibo-fnd-utl-av:explanatoryNote>a control owner (i.e., control person, per SEC regulations) may have some means or right that allows them to exercise control over board composition, other than through proxy assignment or vote.  Not all control persons have this facility, as it is not inherent to having a significant (for example, 20 percent or more) ownership stake.</fibo-fnd-utl-av:explanatoryNote>
+		<fibo-fnd-utl-av:explanatoryNote>A control owner (i.e., control person, per SEC regulations) may have some means or right that allows them to exercise control over board composition, other than through proxy assignment or vote.  Not all control persons have this facility, as it is not inherent to having a significant (for example, 20 percent or more) ownership stake.</fibo-fnd-utl-av:explanatoryNote>
```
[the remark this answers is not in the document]
ElisaKendall: I want to address this and add more richness in a subsequent issue.  I haven't touched this, just the capitalization of the first letter of the note.  Let's defer this to a follow up issue.  Feel free to raise one if you like.

BE/OwnershipAndControl/Executives.rdf
```diff
@@ -127,25 +127,81 @@
...
-		<fibo-fnd-utl-av:definitionOrigin rdf:datatype="&xsd;anyURI">http://www.businessdictionary.com/definition/articles-of-association.html</fibo-fnd-utl-av:definitionOrigin>
+		<skos:definition>contract (1) between members (stockholders, subscribers) and the organization and (2) among the members themselves, that sets out the rights and duties of directors and stockholders</skos:definition>
```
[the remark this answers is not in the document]
ElisaKendall: Revised and eliminated the explanatory notes to abridge several online legal dictionaries.

BE/OwnershipAndControl/Executives.rdf
```diff
@@ -127,25 +127,81 @@
...
+		<fibo-fnd-utl-av:explanatoryNote>Certain statutory clauses (such as those dealing with allotment, transfer, and forfeiture of shares) must be included; the other clauses are chosen by the stockholders to make up the bylaws of the organization. A court, however, may declare a clause ultra vires if it is deemed unfair, unlawful, or unreasonable. A copy of the articles is lodged with the appropriate authority such as the registrar of companies. Articles are public documents and may be inspected by anyone (usually on payment of a fee) either at the premises of the organization or at the registrar&apos;s office. Lenders to the organization take special interest in its provisions that impose a ceiling on the borrowings beyond which the organization&apos;s management must get shareholders&apos; approval before taking on more debt.</fibo-fnd-utl-av:explanatoryNote>
 		<fibo-fnd-utl-av:synonym>articles of association</fibo-fnd-utl-av:synonym>
 	</owl:Class>
 	
 	<owl:Class rdf:about="&fibo-be-oac-exec;Auditor">
-		<rdfs:subClassOf rdf:resource="&fibo-be-oac-exec;ResponsibleParty"/>
+		<rdfs:subClassOf rdf:resource="&fibo-be-oac-exec;AuthorizedParty"/>
```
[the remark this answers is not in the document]
ElisaKendall: The intent was B, but we can add A in a subsequent issue. This was just an initial change to better integrate the concept of an auditor.  It also doesn't address the link between an individual auditor and an outside firm elected as the auditing firm for an organization via annual meeting motions/voting, which is another gap.  Feel free to raise an issue to improve on this.

BE/OwnershipAndControl/Executives.rdf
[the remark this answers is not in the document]
ElisaKendall: At WF (and at at least one other bank that is a FIBO user), integrating information about the relationships between customers is very important.  But what their internal models and data sources do is relate roles to one another, rather than independent people to one another.  Authorization, is critical for a number of cases, including but not limited to role-based access to various resources (not limited to the security issues around that), access to restricted information related to a customer, such as information subject to privacy concerns, etc.  Having a consistent pattern for these things is really important, and so leveraging the lattice pattern allows us to ensure that consistency.  Control relations in FIBO are still a mess.  Ownership is in much better shape now, though we probably need to do a bit more there.  But I need time to really test that, and so I don't want to attempt to get something in for this upcoming release (i.e., Sunday).

BE/OwnershipAndControl/Executives.rdf
```diff
@@ -127,25 +127,81 @@
...
+			</owl:Restriction>
+		</rdfs:subClassOf>
+		<rdfs:subClassOf>
+			<owl:Restriction>
+				<owl:onProperty rdf:resource="&fibo-be-oac-exec;isAuthorizedThrough"/>
+				<owl:someValuesFrom rdf:resource="&fibo-be-oac-exec;Authorization"/>
+			</owl:Restriction>
+		</rdfs:subClassOf>
+		<rdfs:label>authorized party</rdfs:label>
+		<skos:definition>party that has been given the responsibility to act on behalf of another party under some set of guidelines</skos:definition>
```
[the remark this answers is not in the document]
ElisaKendall: They are acting on behalf of the shareholders and regulators.  Guidance depends on the nature of what they are authorized to do.  Depending on the role, there may be licensing required, for example, which is the case for a registered investment advisor and for an outside auditor.  There may be limits to the kinds of information that an authorized party has access to.  There may be limits on the kinds of agreements someone with a power of attorney is authorized to make.  Some of these guidelines or rules will be codified in either a contract or other legal document, or in some policy or something else, which I would represent using restrictions on more specific roles within an organizational context.  This is too high a level to do that.

BE/OwnershipAndControl/Executives.rdf
```diff
@@ -127,25 +127,81 @@
...
+			<owl:Restriction>
+				<owl:onProperty rdf:resource="&fibo-be-oac-exec;authorizesThrough"/>
+				<owl:someValuesFrom rdf:resource="&fibo-be-oac-exec;Authorization"/>
+			</owl:Restriction>
+		</rdfs:subClassOf>
+		<rdfs:label>authorizing party</rdfs:label>
+		<skos:definition>party that delegates some role, authority, or control to another party</skos:definition>
 	</owl:Class>
 	
 	<owl:Class rdf:about="&fibo-be-oac-exec;BoardCapacity">
```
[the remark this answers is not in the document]
ElisaKendall: I have more issues with board membership that I want to address in a separate issue, which will include killing this off.

BE/OwnershipAndControl/Executives.rdf
ElisaKendall: I'll kill the note and add the synonym.

ElisaKendall: I agree that they are distinct from the articles of incorporation, at least in the US

BE/OwnershipAndControl/Executives.rdf
ElisaKendall: The concept is used in the plural in the US - not sure about elsewhere, but the legal document (in the singular) that incorporates the various definitions and guidelines is called "bylaws" in the US - I have the Sandpiper document somewhere that does that, following a template our attorneys provided.

BE/OwnershipAndControl/Executives.rdf
ElisaKendall: fixed - cut and paste error, thanks

BE/OwnershipAndControl/Executives.rdf
ElisaKendall: I thought about deleting this ... in the US the term is corporate law, which refers both to the body of laws and the practice area.  Let's leave it for now and address it in a separate issue, as needed.  I added the synonym in an earlier change to help clarify, because the term 'company law' is the UK term, but didn't remove it.

BE/OwnershipAndControl/Executives.rdf
ElisaKendall: I think there was a slash between the two that the serializer may have removed.  But the distinction isn't one I am familiar with - a corporate secretary is usually an employee rather than a director, but I don't know if there is a legal prohibition from being both.

ElisaKendall: I removed most of that part of the note - our corporate secretary was our attorney for Sandpiper - I think the same is true for OMG, in fact.  The note would have ruled that out.  Maybe that's a jurisdiction specific thing?

BE/OwnershipAndControl/Executives.rdf
ElisaKendall: So the way legally delegated authority is defined, that's the person or organization that is the authority, and there is a restriction saying that they have (possess) delegated legal authority.  I know it's interesting language, but conceptually what's there is correct.  If you can think of a better name after reviewing this in Protege or another tool rather than in the text of the ontology, after running a reasoner, it will make more sense.

BE/OwnershipAndControl/Executives.rdf
ElisaKendall: I'll fix it in the issue I want to raise around board membership once this one is put to bed.

BE/OwnershipAndControl/Executives.rdf
ElisaKendall: It is, and works logically - actsOn is the proper relationship that is the parent in the lattice.  Hard to find a name that fits every case, but linguistically authorizing someone to do something has a primary role of 'actor'.

BE/OwnershipAndControl/Executives.rdf
ElisaKendall: fixed and changed back to party-in-role, since I think one elects a slate of directors at once in some cases, and individual directors in other cases

BE/OwnershipAndControl/Executives.rdf
ElisaKendall: Fixed the wording and this is important for a number of banking role-related use cases - for power of attorney, for signing authority on an account, etc.

BE/OwnershipAndControl/Executives.rdf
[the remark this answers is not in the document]
rivettp: I don't disagree with what you both wrote but:
a) what we agreed as a policy was to use the pattern related to the specific modeling scenario rather than based on consistency among many
b) Customer is not a relative thing for this authorization situation but for the customer relationship. You do not have a relative thing purely for Authorizer but reuse the different Customer relative thing. Fine but not something we've documented any sort of pattern for. This is NOT the 5 object (or even the 3 object) pattern, but something different. I tried to catalog different potential patterns (with diagrams and everything) for the FLT but was roadblocked be people "needing more time to review"
c) IMO we cannot accept as justification undocumented use cases/scenarios which are confidential/secret! Either document them properly or don't use them as justification. How can anyone challenge whether what's proposed is the best model for the use case if it's not transparent? You might as well just say: "trust me, I know what is best because I'm omniscient".

BE/OwnershipAndControl/Executives.rdf
rivettp: You didn't answer my question about how this would ever be used and whether each (voting?) Shareholder would be linked? What if one shareholder voted No or Abstained or gave their Proxy to the company secretary? Not that I think FIBO should be used for vote tallying: what is the point of this property?

BE/OwnershipAndControl/Executives.rdf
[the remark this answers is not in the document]
rivettp: Don't get me wrong, I don't think we need a) - FIBO is not for a directory of qualified auditors. But we do need to make clearer that the intent is b)

Issue #1073, opened 2020-07-02
Eliminate remaining dependencies on the Values ontology and remove it


ElisaKendall:
There are only two elements in the values ontology that any provisional ontology depends on. One of those, percentage value, is already defined in the Analytics ontology. The other, basis points value, if it is actually required, should be defined in IND, likely in the interest rates ontology.

PR #1074, opened 2020-07-02
FND-304 - Eliminate remaining dependencies on the Values ontology and remove it

ElisaKendall:

## Description

Replaced all occurrences of PercentageValue from the provisional Values ontology with Percentage from the Analytics ontology and eliminated the provisional ontology

Fixes: #1073 / FND-304

## Checklist:


ElisaKendall:
@rivettp - the ontology as it stood was incomplete, and the concept of a percentage value was much more well defined in Analytics.  I advocated against its inclusion when the FLT discussed it, but deferred to Cory and Mike Bennett who were charged with making it useful.  That hasn't happened, and the only provisional ontologies that used it at all used percentage value, which again is defined elsewhere.  The only possible datatype we've noticed we might require at some stage for our international users is one that is a subtype of rdfs:Literal that allows one to use either an xsd:string or rdf:langString, but that hasn't come up as instead one can use rdfs:label with multilingual tags, and same for skos:definition or other annotations.

Reified data values -might- be useful in some applications, for mapping data across systems for example, but it hasn't come up in any of our use cases to date.  And the issue with numbers in particular is that one would have to give them an IRI taking this approach.

Issue #1084, opened 2020-07-07
Eliminate unnecessary unions in restrictions in the Occurrences ontology


ElisaKendall:
These unions in restrictions on ad hoc schedule entry, regular schedule, and schedule stub are redundant and add unnecessary complexity, making the ontology harder to understand and increasing reasoning complexity.

This issue relates to #1043.

Issue #1043, opened 2020-06-21
Does FIBO saturate? (OWL profiles)


bblfish:
I read the very helpful 2012 article by Markus Krötzsch [OWL 2 Profiles: An Introduction to Lightweight Ontology Languages](https://link.springer.com/chapter/10.1007/978-3-642-33158-9_4) and was wondering if FIBO aims at a particular profile, eg. OWL DL, RL or QL or some newer variant? 

Reasoning seems to saturate with Hermit in Protégé. (it seems to finish after at least 6 hours of reasoning). I could not tell with Stardog as that uses query rewriting to answer questions and took forever to answer some questions (especially DESCRIBE ones when I set DL reasoning on).

Is FIBO engineered to satisfy a particular reasoning profile? I am especially interested to know if all the consequences can be deduced from FIBO, or FIBO +  instance data.

bblfish:
I have been exploring FIBO with the OWL-API using [ammonite](http://ammonite.io/), which I highly recommend. It would allow the repository to write scripts to test any commit made to it. I will later try to publish a git repo with the scripts I am using.

Loading the full ontology from https://spec.edmcouncil.org/fibo/ontology/master/2020Q1/LoadFIBOProd.rdf which loads 131 ontologies, I find that most of the ontologies only fit OWL-FULL.
But that is because there are minor errors here and there that would be easy to fix to bring it down to OWL-DL.

Looking at OWL-RL I find a lot of errors such as the following, 

```
Use of non-superclass expression in position that requires a superclass expression:
ObjectUnionOf(
   ObjectExactCardinality(1 <https://spec.edmcouncil.org/fibo/ontology/FND/Relations/Relations/comprises> <https://spec.edmcouncil.org/fibo/ontology/FND/Places/Addresses/SupplementalAddressIndicatorOrUnit>)
   DataExactCardinality(1 <https://www.omg.org/spec/LCC/Languages/LanguageRepresentation/hasTag> rdfs:Literal)
)
[
  SubClassOf(
     <https://spec.edmcouncil.org/fibo/ontology/FND/Places/Addresses/SupplementalAddressComponent>
     ObjectUnionOf(
         ObjectExactCardinality(1 <https://spec.edmcouncil.org/fibo/ontology/FND/Relations/Relations/comprises> <https://spec.edmcouncil.org/fibo/ontology/FND/Places/Addresses/SupplementalAddressIndicatorOrUnit>)
[...]

PR #1085, opened 2020-07-07
FND-307 - Eliminate unnecessary unions in restrictions in the Occurrences ontology

ElisaKendall:

## Description

Eliminate unnecessary unions in restrictions in occurrences; note that there is one remaining union in a restriction in this ontology, on the occurrence class, which is definitional and thus retained.

Fixes: #1084 / FND-307; also relates to #1043 

## Checklist:

### DIFF
--- a/FND/DatesAndTimes/Occurrences.rdf
+++ b/FND/DatesAndTimes/Occurrences.rdf
@@ -59,10 +59,10 @@
 		<owl:imports rdf:resource="https://spec.edmcouncil.org/fibo/ontology/FND/Utilities/AnnotationVocabulary/"/>
 		<owl:imports rdf:resource="https://www.omg.org/spec/LCC/Countries/CountryRepresentation/"/>
 		<owl:imports rdf:resource="https://www.omg.org/spec/LCC/Languages/LanguageRepresentation/"/>
-		<owl:versionIRI rdf:resource="https://spec.edmcouncil.org/fibo/ontology/FND/20190201/DatesAndTimes/Occurrences/"/>
+		<owl:versionIRI rdf:resource="https://spec.edmcouncil.org/fibo/ontology/FND/2020701/DatesAndTimes/Occurrences/"/>
 		<skos:changeNote>The https://spec.edmcouncil.org/fibo/ontology/FND/20141101/DatesAndTimes/Occurrences/ version of this ontology was revised to address the issue resolutions in the FIBO 2.0 RFC, primarily to add properties that are relevant to the inputs and outputs from processes, events, systems and the like.</skos:changeNote>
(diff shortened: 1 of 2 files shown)


FND/DatesAndTimes/Occurrences.rdf
```diff
@@ -77,39 +77,35 @@ They are modularized this way to minimize the ontological committments that are
...
-			</owl:Class>
+			<owl:Restriction>
+				<owl:onProperty rdf:resource="&fibo-fnd-dt-oc;hasOccurrence"/>
+				<owl:onClass rdf:resource="&fibo-fnd-dt-oc;Occurrence"/>
+				<owl:minQualifiedCardinality rdf:datatype="&xsd;nonNegativeInteger">0</owl:minQualifiedCardinality>
+			</owl:Restriction>
+		</rdfs:subClassOf>
+		<rdfs:subClassOf>
+			<owl:Restriction>
+				<owl:onProperty rdf:resource="&fibo-fnd-rel-rel;comprises"/>
```
rivettp, line 88: "comprises" seems a very odd property to use here: in no way is the Kind a part of the ScheduleEntry. 

However is it not the Occurrence rather than the ScheduleEntry that has the Kind? And that would be a restriction on the Occurrence - which we already have. Though that again uses an odd property "exemplifies". I would have thought "isClassifiedBy" would be better.
Or is this an attempt to *constrain* all the Occurrences in a given schedule to have the same Kind? If so that should be a restriction on the Schedule class with a cascaded restriction used (on the members of the Schedule). 

Bottom line - I propose to remove the Kind restriction on this, and all other xScheduleEntry classes, and replace the one on Occurrence to use isClassifiedBy instead of exemplifies.

ElisaKendall: So, the only change I made was to eliminate the union, I didn't change the restrictions themselves from what Mark created originally.  The intent, I believe, was to ensure that for a given regular schedule, there is consistency in the kinds of events that occur on a regular basis - the definition of 'regular'.  For example, a regular schedule of payments would have all and only occurrence kinds that are payments.

It's optional at the highest level, so that a scheduler could decide whether a given regular or ad hoc schedule includes only one kind of event, for example, a payment for a fixed rate loan, or more than one kind of event, say, a payment and a rate reset for a variable rate loan.  A given schedule entry should reflect only one kind of occurrence, though.  There are other issues with the restrictions on schedule entry, though - it inherits a combined date and time from dated collection constituent, which seems appropriate and should not be optional, but there are other optional restrictions at the higher level on date or date time that I think are unnecessary and lilkely confusing, so I'll fix that.

To retain the idea that Mark incorporated, that a regular schedule consists of events of a specific kind, or one or more specific kinds, seems important.  Whether comprises is the right verb for that is another matter, but comprises is a synonom of consists of, so I'm hesitant to remove it altogether.  A payment schedule wouldn't include events such as trades or settlements, only payments and possibly rate resets or rate assessments of some pre-determined kind.

Issue #1095, opened 2020-07-11
The domain of of several properties should not be 'Reference', which is too narrow


ElisaKendall:
Other things can be defining concepts, so the domain declaration for 'defines' is too restrictive. The same issue is true of appliesTo, characterizes, and hasArgument.  Some of their definitions need to be made ISO 704 compliant as well.

CoryCasanave:
I hope you replace the domain with something, every property is about something more specific than “thing”

ElisaKendall:
@CoryCasanave I think in these cases that would be tough - they are really general properties, although there may be a class that would work for characterizes, but I'd need to do quite a bit of testing to ensure that doesn't have unintended consequences, which the original domain(s) did for these.

rivettp:
It makes me question what is the value of such general properties? When would you ever query ?s :characterizes x? You'd always use a more specific property. Or filter on the class of ?s. The only argument I've ever heard is that it makes it easier to navigate in limited tools like Protege. Even then I'd disagree - how is someone to know whether to look under classifies or characterizes?

ElisaKendall:
For characterizes, I think the domain could be Aspect - doing some research to see if that will work, but that change has to be made in Analytics, not in Relations, to avoid circular dependencies.

PR #1096, opened 2020-07-11
FND-306 - The domain of of several properties should not be 'Reference', which is too narrow

ElisaKendall:

## Description

Removed 'Reference' as the domain of several properties; eliminated a redundant 'evaluates' property from Analytics that was not used anywhere; adjusted definitions for a few properties in Relations that were not ISO 704 compliant

Fixes: #1095 / [FND-306](https://jira.edmcouncil.org/browse/FND-306)

## Checklist:

### DIFF
--- a/FND/Relations/Relations.rdf
+++ b/FND/Relations/Relations.rdf
@@ -63,7 +63,7 @@ the FIBO FND RFC, was revised in advance of the September 2013 New Brunswick, NJ
 		<skos:changeNote>The https://spec.edmcouncil.org/fibo/ontology/FND/20190501/Relations/Relations.rdf version of this ontology was modified to rename (migrate) the hasDefinition property to isDefinedIn.</skos:changeNote>
 		<skos:changeNote>The https://spec.edmcouncil.org/fibo/ontology/FND/20190701/Relations/Relations.rdf version of this ontology was modified to eliminate deprecated elements.</skos:changeNote>
 		<skos:changeNote>The https://spec.edmcouncil.org/fibo/ontology/FND/20190901/Relations/Relations.rdf version of this ontology was modified to eliminate duplication with concepts in LCC and remove the unused hasDispositionDate property, whose semantics were unclear at best.</skos:changeNote>
-		<skos:changeNote>The https://spec.edmcouncil.org/fibo/ontology/FND/20200301/Relations/Relations.rdf version of this ontology was modified to move hasAcquisitionDate to Financial Dates to improve usability and simplify reasoning, to eliminate circular definitions to deprecate fibo-fnd-rel-rel;hasTag in favor of the simpler LCC property of the same name, and to loosen domain restrictions on some properties which were too narrowly specified.</skos:changeNote>
+		<skos:changeNote>The https://spec.edmcouncil.org/fibo/ontology/FND/20200301/Relations/Relations.rdf version of this ontology was modified to move hasAcquisitionDate to Financial Dates to improve usability and simplify reasoning, to eliminate circular definitions to deprecate fibo-fnd-rel-rel;hasTag in favor of the simpler LCC property of the same name, to loosen domain restrictions on some properties which were too narrowly specified, and to add two new properties, describes and isDescribedBy, which are more appropriate for certain cases where we were using characterizes and isCharacterizedBy.</skos:changeNote>
 		<fibo-fnd-utl-av:hasMaturityLevel rdf:resource="&fibo-fnd-utl-av;Release"/>
 	</owl:Ontology>
 	
@@ -100,10 +100,9 @@ the FIBO FND RFC, was revised in advance of the September 2013 New Brunswick, NJ
 	</owl:ObjectProperty>
 	
 	<owl:ObjectProperty rdf:about="&fibo-fnd-rel-rel;characterizes">
-		<rdfs:subPropertyOf rdf:resource="&fibo-fnd-rel-rel;refersTo"/>
+		<rdfs:subPropertyOf rdf:resource="&fibo-fnd-rel-rel;appliesTo"/>
 		<rdfs:label>characterizes</rdfs:label>
-		<skos:definition>describes the nature, features or quality of</skos:definition>
-		<fibo-fnd-utl-av:adaptedFrom rdf:datatype="&xsd;anyURI">https://www.thefreedictionary.com/characterized</fibo-fnd-utl-av:adaptedFrom>
+		<skos:definition>specifies a discriminating feature or quality of</skos:definition>
 	</owl:ObjectProperty>
 	
 	<owl:ObjectProperty rdf:about="&fibo-fnd-rel-rel;comprises">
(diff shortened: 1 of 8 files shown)


FND/Relations/Relations.rdf
```diff
@@ -81,8 +81,7 @@ the FIBO FND RFC, was revised in advance of the September 2013 New Brunswick, NJ
 	<owl:ObjectProperty rdf:about="&fibo-fnd-rel-rel;appliesTo">
 		<rdfs:subPropertyOf rdf:resource="&fibo-fnd-rel-rel;refersTo"/>
 		<rdfs:label>applies to</rdfs:label>
-		<rdfs:domain rdf:resource="&fibo-fnd-rel-rel;Reference"/>
-		<skos:definition>a relation indicating something that is pertinent or relevant to the concept</skos:definition>
+		<skos:definition>is pertinent, suitable, or relevant to</skos:definition>
```
rivettp, line 84: too broad compared to its actual usage in FIBO

FND/Relations/Relations.rdf
```diff
@@ -103,8 +102,7 @@ the FIBO FND RFC, was revised in advance of the September 2013 New Brunswick, NJ
 	<owl:ObjectProperty rdf:about="&fibo-fnd-rel-rel;characterizes">
 		<rdfs:subPropertyOf rdf:resource="&fibo-fnd-rel-rel;refersTo"/>
 		<rdfs:label>characterizes</rdfs:label>
-		<rdfs:domain rdf:resource="&fibo-fnd-rel-rel;Reference"/>
-		<skos:definition>a feature or quality that distinguishes something from something else</skos:definition>
+		<skos:definition>describes the nature, features or quality of</skos:definition>
```
rivettp, line 105: Too broad to be useful. The giveaway is the "or" in the definition - maybe we should have a hygiene test for that!

It's impossible from these definitions to decide whether to use appliesTo or characterizes. Seriously.
We have Catalog characterizes Product; why would it not be appliesTo?
Given that characterizes is used with Aspect (which is at least coherent) then, simce Catalog is nothing at all like an Aspect, that Catalog usage is wrong.

qualifies is subPropertyOf characterizes but with quite different, rather than more specific meaning "limits, constrains, or refines". Again not coherent.

Another usage is that a specific CodeScheme characterizes a CodeRepository. I wouldn't say that it "describes the quality of" it. Again not coherent.

FND/Relations/Relations.rdf
```diff
@@ -132,14 +130,13 @@ the FIBO FND RFC, was revised in advance of the September 2013 New Brunswick, NJ
...
 		<rdfs:label>defines</rdfs:label>
-		<rdfs:domain rdf:resource="&fibo-fnd-rel-rel;Reference"/>
 		<skos:definition>determines or identifies the essential qualities or meaning of, discovers and sets forth the meaning of, fixes or marks the limits of, demarcates</skos:definition>
 	</owl:ObjectProperty>
 	
 	<owl:ObjectProperty rdf:about="&fibo-fnd-rel-rel;designates">
 		<rdfs:label>designates</rdfs:label>
 		<rdfs:domain rdf:resource="&fibo-fnd-aap-agt;AutonomousAgent"/>
-		<skos:definition>to name something officially or appoint someone to a position officially</skos:definition>
+		<skos:definition>names something officially or appoints someone to a position officially</skos:definition>
```
rivettp, line 139: No no no.
These are entirely different semantics. They might use the same English word but must not use the same semantic property.

FND/Utilities/Analytics.rdf
```diff
@@ -101,6 +101,14 @@
 		<owl:equivalentClass rdf:resource="&fibo-fnd-utl-alx;Ratio"/>
 	</owl:Class>
 	
+	<owl:ObjectProperty rdf:about="&fibo-fnd-rel-rel;characterizes">
+		<rdfs:domain rdf:resource="&fibo-fnd-utl-alx;Aspect"/>
+	</owl:ObjectProperty>
+	
+	<owl:ObjectProperty rdf:about="&fibo-fnd-rel-rel;isCharacterizedBy">
+		<rdfs:range rdf:resource="&fibo-fnd-utl-alx;Aspect"/>
```
rivettp, line 109: we have AddressComponent as a subclass of Aspect, which I think is wrong: these restrictions do not work at all well for that.

FND/Relations/Relations.rdf
ElisaKendall: There are actually 77 uses of the property appliesTo in FIBO.  Some of those uses can be replaced with a more specific property, I agree, but I don't know if there is a replacement that is more specific available for every one of them.  I would rather not make a change to this until we have (1) replaced some of its uses with appropriately more specific properties, and (2) done further research on the remainder to decide whether there is a better definition that could reflect its usage for all of them.  The main change here was to eliminate "a relation indicating something that", which is overly verbose and doesn't meet our ISO 704 compliant approach to definitions.  I didn't change the meaning at all.

FND/Relations/Relations.rdf
ElisaKendall: It is not too broad to be useful - that's the definition of the term 'characterizes'.  I agree that some of the places where it is used should be corrected, such as in the case of a catalog, and will make that change.  Applies to is fine for that case and for the code scheme case, which I will revise accordingly.

FND/Relations/Relations.rdf
ElisaKendall: Ok - I've limited this to naming something offucially, and appoints, as a subproperty makes more sense now

FND/Utilities/Analytics.rdf
ElisaKendall: fixed - made it a constituent instead

FND/Relations/Relations.rdf
rivettp: My problem here is less with the property itself, but that the proposed definition does not reflect how it's  used, and should be used.

FND/Relations/Relations.rdf
rivettp: No, how is "appoints" a subproperty of "names something officially"? An example of the latter might be to name the UK as "The United Kingdom of Grate Britain and Northern Ireland". 
Another reason it does not work for naming is that, if tracking the Authorizer (which this does as the domain of the property), and the object of the example triple is the country, then there seems no way to represent the official name that has been designated.

The other alternative "appoints someone" would make more sense for consistency, since the only uses of designates are as super property of a) appoints and b) delegatesControlTo.

My main concern with using "designates" in this way is that we use "designation" inconsistently to mean both "appointment" (consistent) and naming (inconsistent) - for example we have property preferredDesignation.
In summary I think we should stick to something like "appoints" here and remove the naming-related uses of "designation" as a separate issue.

FND/Relations/Relations.rdf
ElisaKendall: Raised FND-311 to review uses of appliesTo; added FND-312 to replace hasDescription with isDescribedBy

FND/Relations/Relations.rdf
ElisaKendall: Ok - I will revise the definition of designates to be appoints someone, and then eliminate the subproperty 'appoints', since it is redundant given this change.

Issue #1081, opened 2020-07-07
Extending hygiene tests set - no unused imports


mereolog:
https://github.com/edmcouncil/fibo/blob/master/ONTOLOGY_GUIDE.md#minimal-compliance-hygiene-tests suggests a test to the effect that no ontology is imported (by a given ontology) unless it is explicitly referenced in the importing ontology.

I am not sure what the exact content of this test is.
Looking at the code there I can see two options:
1.  no ontology is imported (by a given ontology) unless the importing ontology has a triple with an entity (i.e., class, property, or individual) from the imported ontology.
a. the imported entity may occupy any position in the triple: subject's, property's, or object's.
2. no ontology is imported (by a given ontology) unless the importing ontology has a triple where:
a. the property is rdfs:isDefinedBy
b. the object is the imported ontology itself.

Of course both interpretations can be intended, but then this test should probably be split into two.
Also the test may have been designed to achieve something else. 

I was wondering whether someone could inform me about the intended interpretation of this test.

ElisaKendall:
I think the main issue is 1. above.  We really want to make sure that if something in an external ontology is referenced by any s, p, o, that the reference resolves, but that we don't import anything 'extra'.  I'm certain there are still cases of this - mainly in provisional ontologies where we've referenced things, then elements were moved around, and we forgot to remove the no longer needed imports.  There are also cases where in the original generation of the OWL, imports were added that were never used.  In the released ontologies there should be fewer cases, if any.

mereolog:
Then this query should do the trick - although it is painfully slow :), sp perhaps there is a space for optimisation.
```
prefix owl:   <http://www.w3.org/2002/07/owl#>
prefix rdf:   <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
prefix rdfs:  <http://www.w3.org/2000/01/rdf-schema#>

SELECT DISTINCT ?error
WHERE 
{
    ?importingOntology owl:imports ?importedOntology.
    FILTER (CONTAINS(str(?importingOntology), "edmcouncil"))
    FILTER NOT EXISTS 
    {
        ?s ?p ?o .
        FILTER (?p NOT IN (owl:imports))
        FILTER (CONTAINS(str(?s), "edmcouncil") || CONTAINS(str(?p), "edmcouncil") || CONTAINS(str(?o), "edmcouncil"))
        FILTER (CONTAINS(CONCAT(str(?s), str(?p), str(?o)),str(?importedOntology))).
    }
[...]

PR #1116, opened 2020-07-22
Extending hygiene tests set - no unused imports

mereolog:
Signed-off-by: Pawel Garbacz <pawel.garbacz@makolab.com>

## Description

This test is to find unused imports in the whole (aggregated) FIBO ontology.

Fixes: #1081

## Checklist:

- [X] I'm familiar with the [FIBO developer quide](../CONTRIBUTING.md#contributing-to-the-fibo-code). My contribution meets all the requirements described there.
- [X] My contribution follows the [principles of best practices for FIBO](../ONTOLOGY_GUIDE.md).
- [X] My changes have been reconciled with latest master and no merge conflicts remain.
- [X] This PR is related to exactly one issue. The issue is referenced by using a GitHub keyword such as "fixes", "closes", or "resolves".
- [X] Hygiene tests have been applied by a PR with "(WIP)" in title.
- [X] The issue has been tested locally using a reasoner (for ontology changes).

### DIFF
--- a/BE/GovernmentEntities/GovernmentEntities.rdf
+++ b/BE/GovernmentEntities/GovernmentEntities.rdf
@@ -78,13 +78,14 @@
 		<owl:imports rdf:resource="https://spec.edmcouncil.org/fibo/ontology/FND/Utilities/AnnotationVocabulary/"/>
 		<owl:imports rdf:resource="https://www.omg.org/spec/LCC/Countries/CountryRepresentation/"/>
 		<owl:imports rdf:resource="https://www.omg.org/spec/LCC/Languages/LanguageRepresentation/"/>
-		<owl:versionIRI rdf:resource="https://spec.edmcouncil.org/fibo/ontology/BE/20200301/GovernmentEntities/GovernmentEntities/"/>
+		<owl:versionIRI rdf:resource="https://spec.edmcouncil.org/fibo/ontology/BE/20200701/GovernmentEntities/GovernmentEntities/"/>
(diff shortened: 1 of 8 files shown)


mereolog:
@rivettp 
As for:
> Hmm I don't see where the triple ?s ?p ?o is linked to the ?importingOntology. Which would actually be hard to do unless you use the quad representation which has a named graph for each ontology.
> 
I can add the following clause to FILTER NOT EXISTS filter:
`FILTER (CONTAINS(str(?s), str(?importingOntology)) || CONTAINS(str(?p), str(?importingOntology)) || CONTAINS(str(?o), str(?importingOntology)))
`

mereolog:
@rivettp 
As for:

> There could be some false matches here (which will be considered a usage whereas it's not) e.g. if one ontology has a URI which is equal to another with an extra word or two added.
> Also if a URI is contained in the string that is the object of a datatype or annotation property.

If I get you right, the following change may a remedy.
Instead of:
`FILTER (CONTAINS(CONCAT(str(?s), str(?p), str(?o)),str(?importedOntology))).`
we could have:
`FILTER (REGEX(str(?s), CONCAT(str(?importedOntology),".+")) || REGEX(str(?p), CONCAT(str(?importedOntology),".+")) || REGEX(str(?o), CONCAT(str(?importedOntology),".+")))`

ElisaKendall:
One thing I am wondering about - there is at least one case where I've explicitly imported an ontology in the latest revisions to loans because that ontology adds restrictions on DebtInstrument that we need for a loan, but there are no direct references to anything.  Would this require that I add some reference to the SEC/Debt/DebtInstruments ontology in the importing ontology to keep that import statement?

mereolog:
@ElisaKendall 
Which ontology did you mean?

Issue #1124, opened 2020-08-05
Additional government entities and jurisdictions are needed to fully cover the countries in Europe


ElisaKendall:
The released ontologies include coverage of what the M49 codes consider NorthernAmerica, but very little for what M49 considers Europe.  These entities and jurisdictions are needed to provide additional organizational and regulatory coverage of Europe, to support additional examples, and to enable FIBO users to extend these entities or reuse them as needed in numerous use cases, including those related to risk analytics.

PR #1125, opened 2020-08-05
BE-224 - Additional government entities and jurisdictions are needed to fully cover the countries in Europe

ElisaKendall:

## Description

This issue resolution brings in polities, governments, and jurisdictions for all of the countries considered part of Europe in the U.N. M49 codes.  They are modularized regionally, again based on the approach that the M49 codes take, including Northern Europe, Southern Europe, Western Europe, and Eastern Europe.  Only those countries identified in the ISO 3166-1 codes are covered here, which excludes additional entities for European country subdivisions.   These can be added, as needed, over time, but at a minimum, coverage for each of the independent countries is important.

Fixes: #1124 / BE-224

## Checklist:


MikeHypercube:
Given the addition of these governments and countries classified according to their physical region (UN), what remains as the purpose of the also-imported ontology that is EU entities and jurisdictions? Is that ontology now empty (I see no redline for it) or does it define the EU jurisdiction alone (singular, but the name suggests multiple)? Or are some entities defined in more than one such ontology now? e.g. all Western Europe apart from Switzerland and a couple of principalities are both Western Europe and EU countries.  

As it stands it would not be clear to a user where to look for e.g. Austria. Would that be an issue for users?

ElisaKendall:
> Given the addition of these governments and countries classified according to their physical region (UN), what remains as the purpose of the also-imported ontology that is EU entities and jurisdictions? Is that ontology now empty (I see no redline for it) or does it define the EU jurisdiction alone (singular, but the name suggests multiple)? Or are some entities defined in more than one such ontology now? e.g. all Western Europe apart from Switzerland and a couple of principalities are both Western Europe and EU countries.
> 
> As it stands it would not be clear to a user where to look for e.g. Austria. Would that be an issue for users?

Hi Mike,

The EU ontology is intended to provide the definition of the European Union, and once we have additional concepts, such as monetary zone and trade bloc defined, we will add the Eurozone and other trading zones that are specific to Europe to that ontology.

So, any country that is a member of the EU would be included in the EU ontology, though I think right now we have included the countries and I think it needs to be the governments.  There is an outstanding issue against that ontology to remove the UK, and we could use that to replace countries with governments if that's the right thing to do.

Elisa

MikeHypercube:
Paragraph 2 of your response leaves me even more confused. How  can the EU ontology contain any country that is in the EU, when every such country is already in Western Europe, Eastern Europe or Southern Europe? 

Otherwise you would end up with 2 of Austria, 2 of France, etc x26

And yes, the UK has left so it has to come out of the EU one - but is already correctly included in Western Europe.

ElisaKendall:
> Paragraph 2 of your response leaves me even more confused. How can the EU ontology contain any country that is in the EU, when every such country is already in Western Europe, Eastern Europe or Southern Europe?
> 
> Otherwise you would end up with 2 of Austria, 2 of France, etc x26
> 
> And yes, the UK has left so it has to come out of the EU one - but is already correctly included in Western Europe.

Hi Mike,

It doesn't redefine them - it only references them, for example, using hasReach for jurisdiction.

MikeHypercube:
Now it makes sense - I'd forgotten that Countries themselves are in LCC not FIBO. 

Hopefully the answer to my original question (is this confusing for users) is No, as long as people know that for these concepts they should look in the regional and not the bloc-related ontology with otherwise similar names. I recommend we think about some kind of guidance notes on that so people don't end up putting some SovereignState into the EU ontology (e.g. if and when Serbia becomes an EU member state).

ElisaKendall:
> I think separate individuals for Country and Jurisdiction is, in 99% of the cases where they're 1-to-1, pointless and harmful: it will cause confusion and unwieldiness and put people off FIBO. Someone modeling data is not going to want to consider whether to link to the country or the jurisdiction - it's not a distinction the majority using FIBO need care about (WHERE"S THE USE CASE??) We even get it wrong within FIBO (why for example is a Currency linked to a Country not a Jurisdiction since it represents legal tender?)

I agree that the legal tender should be linked to the **_issuer_** not the country or the jurisdiction.  The issuer is the government. Jurisdiction is quite different from the country - there are many overlapping jurisdictions in many places to consider.  And yes, it's not the primary focus for FIBO, but it is becoming increasingly important for some of the stakeholders, especially for anti-money laundering, KYC, and other use cases.  For risk management, there are a number of use cases that require the governments, jurisdictions, etc.  Citi, WF, JPMC, and others have this need.  But we haven't had time to set up yet another working group to address some of these things. I really disagree with you Pete that the same individual should be used to represent the country and the jurisdiction - in many cases but not all they may appear to be 1-1, but it's not true across the board, and we are providing training via the council that can help users understand the distinction.

Once the government entities and jurisdictions are available, I'll raise an issue to move the currency ontology to FBC, so that we get the pattern right, linking the currencies to the governments that issue them and to the central banks responsible for managing them.  This is a stepwise change, though - it is tedious and time-consuming to get the details right, which cannot be generated.

rivettp:
While we *could* track who "issues" a currency, the current property we have is "isUsedBy" populated from ISO 4217. The definition is somewhat weak in this context "is employed in the process of accomplishing something" but I think the only reasonable interpretation is that the currency can legally be used to buy something - so surely if anything it should be the (legal) jurisdiction e.g. a 50 Euro note could be issued in Germany but it is legal tender in a whole bunch of countries. 

BTW if you really want to track who "issues" a currency surely it's the central bank not the government. Again I don't think it matters - if you think it does let's have the use case. 

My purpose for the above discussion was that forcing the "country vs jurisdiction" decision on people in all cases is just so unfair on people trying to use FIBO. It's usually really subtle as indicated by the fact that even we have not always got it right.

Another example - language. Again we are linking to Country but arguably it should be Jurisdiction at least for the official administrative languages since it's the official legally acceptable language.

Another example is operatesInCountry for Exchange. This is defined as "indicates the country in which the exchange is registered and operates" - since it's related to registration then again arguably this should be Jurisdiction!

And hasCitizenship. Again a legal thing so should have range Jurisdiction? 

BTW you have not provided a use case for the current PR beyond saying "it is becoming increasingly important for some of the stakeholders, especially for anti-money laundering, KYC, and other regulatory related use cases." Let's have the competency questions at least.

I do agree that it's not 1-1 across the board but let's retain the separate individuals for where they're really needed.

dallemang:
I am confused by Pete's comment that Countries and Jurisdictions are in 1-1 correspondence; States and Provinces also have Jurisdictions (as I suppose do Cities), and these overlap with Country Jurisdictions, and have other complex political relationships with them; it seems to me that Jurisdiction is one of the many things that States (provinces, cantons, prefectures, departments, counties) have in common with Countries.

On another front, the comment that include this quote is marked as a "change request" - maybe this is just a weird feature of github, but I can't figure out what change is being requested; is it to remove all jurisdictions, and change all references to them to be references to the country or country subdivision instead?  Or to change the linkage of a Currency from Country to Jurisdiction?  I don't know how to determine if the change has been made, if I don't know what change was requested.   In both of these cases, the change seems to have much larger scope than the original issue, which was simply to include European entities into already existing lists that only had North American entities in it.

rivettp:
@dallemang if a Jurisdiction is 1-1 with a Subregion (State, Province etc) then that Subregion should be multiply classified with Jurisdiction. That makes it a lot easier for people to reference without worrying whether they mean the Subregion or Jurisdiction, which as a I said is a subtle distinction we should not force people to make a decision on.
Any properties related to Jurisdiction would have this unified individual as their subject/object.
To your questions:
- YES I propose to remove MOST jurisdictions, and change all references to them to be references to the country or country subdivision instead
- NO I'd prefer not to change the linkage of a Currency from Country to Jurisdiction BUT it seems we're forced into it if ignore my first proposal and we perpetuate and claim important the non-helpful subtle distinction between the individuals.

I agree that this PR was really about following through more widely on the current approach of separate individuals. But I think that rather than perpetuating the problematic approach we should change this one to add jurisdiction as a further class to the existing Country/Subdivision individuals and have a (separate I guess) PR to do the same with the existing ones, eliminating the Jurisdiction-specific individuals.

ElisaKendall:
@rivettp I totally disagree.  Jurisdiction is a very different concept from Country, even if in some cases they appear to be extensionally the same.  The bigger concern I have is that there are cases, such as the UK, where it is more difficult to model.  England and Wales is a separate subdivision in the current LCC regions from England,  though you've mentioned that these will all disappear.  England is called a country, England and Wales is called a nation by the UK, again in the version we currently have.  Wales alone is considered a devolved government, but England does not have its own devolved government.  The military jurisdiction covers the national government, which does not define itself as being a federated sovereignty as far as I've been able to determine.  The nature of the government for the United Kingdom of Great Britain and Northern Ireland is 'unitary parliamentary constitutional monarchy', nothing about federal.  Many islands in the Caribbean have QE as their monarch but are considered independent countries from the UK.  This makes representing jurisdictions more complicated than you might imagine at first blush.

For FIBO our main concern is regulatory, i.e., who regulates a given institution for what purpose, and whose jurisdiction do they fall under.  This is really critical for regulatory reporting, compliance, and which usury and other laws, including AML, apply.   When there are multiple jurisdictions, which takes precedence and for what purpose?  It will be far easier to adjust things downstream to add overlapping or new jurisdictions if we keep these concepts and the related individuals distinct from one another, rather than typing the LCC individuals as jurisdictions.  

I see this as being akin to the distinction between the morning star and evening star in some ways - both are extensionally the moon, but there are subtle distinctions.  Those are more important poetically than practically, but as we start to do more work related to the jurisdiction in which some securities regulation applies, we will need to be increasingly careful and granular about jurisdictions.  And, modeling them one way in some cases and a different way for others will require not only justification but legal SMEs to ensure that we are correct.  We may get those at some stage, but they are difficult to come by.

rivettp:
You seem to be disagreeing with something I've never proposed...
I have always agreed they should be separate concepts (Classes). If, as you say, they can be coextensional then why not have separate individuals? And I've also said I'm happy using separate individuals where're not coextensional and there is something different we can justify saying about each. Let's not make the 99% of cases unnecessarily painful for a subtle edge case (which we can still handle). The model (T-box) is identical in both cases. It's akin to how you conflate multiple Role individuals and give them multiple classes.

dallemang:
So - in the case where the jurisdiction is the same as the country (or state), do we populate the property hasReach to be reflexive, e.g., 

:US fibo-fnd-law-jur:hasReach :US .
:US a fibo-fnd-law-jur:Jurisdiction , lcc-cr:GeographicRegion .



?

rivettp:
I'd say yes to reflexive hasReach.

ElisaKendall:
We have not done that to date - I would need to see what that does to the reasoning results, especially in cases the UK, where you might have different jurisdictions for different purposes.

rivettp:
Sorry for lack of clarity, I was responding to @dallemang 's question " we populate the property hasReach to be reflexive". I was not proposing that the Property be defined to be reflexive.
By populating it to be reflexive I mean that the property definition remains the same but we add, in FIBO, to take an example,:

   <owl:NamedIndividual rdf:about="&lcc-3166-1;Finland">
      <rdf:type rdf:resource="&fibo-fnd-law-jur;Jurisdiction"/>
      <fibo-fnd-law-jur:hasReach rdf:resource="&lcc-3166-1;Finland" />

LCC already has:
      <rdf:type rdf:resource="&lcc-cr;Country"/> 
       
As an aside, I see in the Jurisdiction class definition that we have the following restriction:
		<rdfs:subClassOf>
			<owl:Restriction>
				<owl:onProperty rdf:resource="&fibo-fnd-rel-rel;isGovernedBy"/>
				<owl:someValuesFrom rdf:resource="&fibo-fnd-law-jur;LegalSystem"/>
			</owl:Restriction>
		</rdfs:subClassOf>
But we're never populating what we're saying is mandatory. Which I think is bad practice. We should make it min zero or remove the property completely along with LegalSystem class.

ElisaKendall:
I'm still against doing what you suggest with respect to jurisdiction, i.e., the Finland example.  There are many different kinds of jurisdictions, jurisdiction of the supreme court or highest national court for some country being the one that we have been addressing so far, and there are many others.  In the US, for example, the Federal Reserve district banks each have different jurisdictions although they are part of an overall federal system; the OCC has national jurisdiction, as well, but may have defined districts that various examiners cover. The Federal Home Loan banks have different jurisdictions that they cover, again national, but with districts that differ from the Fed and from the OCC, and that doesn't include the FDIC or NCUA.  In other words, even the banking regulators with national jurisdiction have set up districts that their divisions cover, and they don't overlap.  Saying that FInland has a jurisdiction of FInland doesn't mean anything.  Saying that the supreme court of Finland has jurisdiction that extends nationally actually means something.

I agree, on the other hand that we should raise an issue to get rid of the property and classes related to legal system that you've identified.  I'll raise a separate issue in FND for that.

Issue #1129, opened 2020-08-25
The equity instruments ontology is missing details needed to represent preferred shares


ElisaKendall:
As a part of the effort undertaken by the SEC FCT to be able to represent securities as classified in the CFI standard, we've uncovered a number of gaps related to preferred shares.  These gaps should be addressed in a top-down fashion, so that equity instruments can be classified along a number of facets, including facets specific to preferred shares.

This issue is about the addition of the details necessary to represent preferred shares, including moving a few general concepts, such as redemption provisions, from debt to the more general financial instruments ontology.

PR #1130, opened 2020-08-25
SEC-73 - The equity instruments ontology is missing details needed to represent preferred shares

ElisaKendall:

## Description

Revise the equities ontology and others on which it depends to incorporate concepts related to preferred shares, including moving a few general concepts, such as redemption provisions, from debt to the more general financial instruments ontology, to support the classification of preferred shares based on the CFI standard, as well as to enable faceted search of preferred shares in implementations.

Fixes: #1129 / SEC-73

## Checklist:


SEC/Equities/EquityInstruments.rdf
```diff
@@ -406,12 +420,73 @@
 		<skos:definition xml:lang="en">position in an equity instrument</skos:definition>
 	</owl:Class>
 	
+	<owl:Class rdf:about="&fibo-sec-eq-eq;EquityRedemptionProvision">
+		<rdfs:subClassOf rdf:resource="&fibo-fbc-fi-fi;RedemptionProvision"/>
+		<rdfs:subClassOf>
+			<owl:Restriction>
+				<owl:onProperty rdf:resource="&fibo-sec-eq-eq;hasMinimumRedemptionPrice"/>
+				<owl:onClass rdf:resource="&fibo-fbc-fi-ip;MarketPrice"/>
```
rivettp, line 428: MarketPrice seems the wrong class here - it has a pricingSource (market) and also a dateTime. Neither of which are appropriate. Why not MonetaryPrice as per the next restriction?

SEC/Equities/EquityInstruments.rdf
```diff
@@ -406,12 +420,73 @@
...
+				<owl:minQualifiedCardinality rdf:datatype="&xsd;nonNegativeInteger">0</owl:minQualifiedCardinality>
+			</owl:Restriction>
+		</rdfs:subClassOf>
+		<rdfs:subClassOf>
+			<owl:Restriction>
+				<owl:onProperty rdf:resource="&fibo-sec-eq-eq;hasRedemptionPremium"/>
+				<owl:onClass rdf:resource="&fibo-fnd-acc-cur;MonetaryPrice"/>
```
rivettp, line 435: Not a percentage?

SEC/Equities/EquityInstruments.rdf
```diff
@@ -406,12 +420,73 @@
...
+				<owl:minQualifiedCardinality rdf:datatype="&xsd;nonNegativeInteger">0</owl:minQualifiedCardinality>
+			</owl:Restriction>
+		</rdfs:subClassOf>
+		<rdfs:subClassOf>
+			<owl:Restriction>
+				<owl:onProperty rdf:resource="&fibo-sec-eq-eq;hasEarliestRedemptionDate"/>
+				<owl:onClass rdf:resource="&fibo-fnd-dt-fd;ExplicitDate"/>
```
rivettp, line 442: Is this always absolute? And not relative to the effective date e.g. "after 2 years"?

SEC/Equities/EquityInstruments.rdf
```diff
@@ -406,12 +420,73 @@
 		<skos:definition xml:lang="en">position in an equity instrument</skos:definition>
 	</owl:Class>
 	
+	<owl:Class rdf:about="&fibo-sec-eq-eq;EquityRedemptionProvision">
```
rivettp, line 423: Should this not be named specific to PreferredEquity? Can you redeem any other type of equity?

SEC/Equities/EquityInstruments.rdf
```diff
@@ -406,12 +420,73 @@
...
+			<owl:Restriction>
+				<owl:onProperty rdf:resource="&fibo-sec-eq-eq;isRedeemableAtIssuerOption"/>
+				<owl:minQualifiedCardinality rdf:datatype="&xsd;nonNegativeInteger">0</owl:minQualifiedCardinality>
+				<owl:onDataRange rdf:resource="&xsd;boolean"/>
+			</owl:Restriction>
+		</rdfs:subClassOf>
+		<rdfs:subClassOf>
+			<owl:Restriction>
+				<owl:onProperty rdf:resource="&fibo-sec-eq-eq;isRedeemableAtShareholderOption"/>
+				<owl:minQualifiedCardinality rdf:datatype="&xsd;nonNegativeInteger">0</owl:minQualifiedCardinality>
```
rivettp, line 456: Can both this and IssuerOption be false? True? Seems like an enum would be tighter than 2 booleans.

SEC/Equities/EquityInstruments.rdf
```diff
@@ -406,12 +420,73 @@
...
+			</owl:Restriction>
+		</rdfs:subClassOf>
+		<rdfs:label xml:lang="en">equity redemption provision</rdfs:label>
+		<skos:definition xml:lang="en">redemption provision that specifies the conditions under which the issuer or shareholder may redeem the shares</skos:definition>
+	</owl:Class>
+	
+	<owl:Class rdf:about="&fibo-sec-eq-eq;ExtendablePreferredShare">
+		<rdfs:subClassOf rdf:resource="&fibo-sec-eq-eq;RedeemablePreferredShare"/>
+		<rdfs:label>extendable preferred share</rdfs:label>
+		<skos:definition>redeemable preferred share whose redemption date can be extended</skos:definition>
```
rivettp, line 467: who can extend it?
RedemptionProvision above has only an earliestRedeemableDate - what other date is there?

SEC/Equities/EquityInstruments.rdf
```diff
@@ -406,12 +420,73 @@
...
+	<owl:Class rdf:about="&fibo-sec-eq-eq;FixedRateDividend">
+		<rdfs:subClassOf rdf:resource="&fibo-sec-eq-eq;PreferredDividend"/>
+		<rdfs:subClassOf>
+			<owl:Restriction>
+				<owl:onProperty rdf:resource="&fibo-sec-eq-eq;hasFixedDividendRate"/>
+				<owl:someValuesFrom rdf:resource="&fibo-fnd-acc-cur;PercentageMonetaryAmount"/>
+			</owl:Restriction>
+		</rdfs:subClassOf>
+		<rdfs:label>fixed rate dividend</rdfs:label>
+		<skos:definition>dividend that provides a specified annual return on the nominal value (and any premium) paid on shares</skos:definition>
```
rivettp, line 486: I don't understand the "(and any premium)". If the fixed rate is 10% and I paid $100 with $10 premium does that mean I get $11?
Separately is the interest compounded or it based strictly on what was originally paid?
Finally there is a huge difference between "nominal value" and "amount paid" on shares! The note seems to acknowledge that but the definition needs fixing.

SEC/Equities/EquityInstruments.rdf
```diff
@@ -572,45 +667,29 @@
...
-				<owl:onProperty rdf:resource="&fibo-sec-eq-eq;hasDistributionMethod"/>
-				<owl:onClass rdf:resource="&fibo-sec-eq-eq;DividendDistributionMethod"/>
-				<owl:maxQualifiedCardinality rdf:datatype="&xsd;nonNegativeInteger">1</owl:maxQualifiedCardinality>
+				<owl:onProperty rdf:resource="&fibo-sec-eq-eq;hasDividend"/>
+				<owl:someValuesFrom rdf:resource="&fibo-sec-eq-eq;AdjustableRateDividend"/>
 			</owl:Restriction>
 		</rdfs:subClassOf>
+		<rdfs:label>preferred share with adjustable rate dividend</rdfs:label>
+		<owl:disjointWith rdf:resource="&fibo-sec-eq-eq;PreferredShareWithFixedRateDividend"/>
+		<skos:definition>preferred share whose dividends will vary with a benchmark, most often a T-bill rate</skos:definition>
```
rivettp, line 680: Need a property to reference the Benchmark; and there will presumably be an adjustment e.g. +0.5%

SEC/Equities/EquityInstruments.rdf
```diff
@@ -627,6 +706,18 @@
 		<skos:definition>dividend that falls under capital gains tax rates that are lower than the income tax rates on unqualified, or ordinary, dividends</skos:definition>
 	</owl:Class>
 	
+	<owl:Class rdf:about="&fibo-sec-eq-eq;RedeemablePreferredShare">
```
rivettp, line 709: Are any of these subclasses of PreferredShare disjoint?

SEC/Equities/EquityInstruments.rdf
```diff
@@ -768,29 +859,37 @@
 		<skos:definition xml:lang="en">grants or bestows ownership via contract</skos:definition>
 	</owl:ObjectProperty>
 	
-	<owl:ObjectProperty rdf:about="&fibo-sec-eq-eq;hasCoupon">
+	<owl:ObjectProperty rdf:about="&fibo-sec-eq-eq;hasAdjustableDividendRate">
+		<rdfs:subPropertyOf rdf:resource="&lcc-lr;has"/>
+		<rdfs:label>has adjustable dividend rate</rdfs:label>
+		<rdfs:domain rdf:resource="&fibo-sec-eq-eq;PreferredDividend"/>
+		<rdfs:range rdf:resource="&fibo-fnd-acc-cur;PercentageMonetaryAmount"/>
```
rivettp, line 866: This is not adequate to represent a formula.

SEC/Equities/EquityInstruments.rdf
```diff
@@ -768,29 +859,37 @@
...
+	</owl:ObjectProperty>
+	
+	<owl:ObjectProperty rdf:about="&fibo-sec-eq-eq;hasDistributionMethod">
 		<rdfs:subPropertyOf rdf:resource="&lcc-lr;has"/>
-		<rdfs:label>has coupon</rdfs:label>
-		<rdfs:domain rdf:resource="&fibo-sec-eq-eq;PreferredShare"/>
+		<rdfs:label>has distribution method</rdfs:label>
+		<rdfs:domain rdf:resource="&fibo-sec-eq-eq;Dividend"/>
+		<rdfs:range rdf:resource="&fibo-sec-eq-eq;DividendDistributionMethod"/>
+		<skos:definition>indicates the method by which dividend payments are to be distributed</skos:definition>
```
rivettp, line 875: don't need "indicates"

SEC/Equities/EquityInstruments.rdf
```diff
@@ -802,12 +901,20 @@
...
+		<rdfs:domain rdf:resource="&fibo-sec-eq-eq;PreferredDividend"/>
+		<rdfs:range rdf:resource="&fibo-fnd-dt-fd;Duration"/>
+		<skos:definition>indicates a period of time after a dividend payment becomes due, before the issuer is subject to penalties</skos:definition>
+	</owl:ObjectProperty>
+	
+	<owl:ObjectProperty rdf:about="&fibo-sec-eq-eq;hasEarliestRedemptionDate">
+		<rdfs:subPropertyOf rdf:resource="&fibo-fnd-dt-fd;hasExplicitDate"/>
+		<rdfs:label>has earliest redemption date</rdfs:label>
+		<rdfs:domain rdf:resource="&fibo-sec-eq-eq;EquityRedemptionProvision"/>
+		<rdfs:range rdf:resource="&fibo-fnd-dt-fd;ExplicitDate"/>
```
rivettp, line 916: Surely should be able to express this as a duration after effective date

SEC/Equities/EquityInstruments.rdf
```diff
@@ -832,12 +947,12 @@
...
-		<rdfs:subPropertyOf rdf:resource="&fibo-fnd-dt-fd;hasDuration"/>
-		<rdfs:label>has grace period</rdfs:label>
-		<rdfs:domain rdf:resource="&fibo-sec-eq-eq;PreferredShareCoupon"/>
-		<rdfs:range rdf:resource="&fibo-fnd-dt-fd;Duration"/>
-		<skos:definition>indicates a period of time after a coupon payment becomes due, before the issuer is subject to penalties</skos:definition>
+	<owl:ObjectProperty rdf:about="&fibo-sec-eq-eq;hasMinimumRedemptionPrice">
+		<rdfs:subPropertyOf rdf:resource="&fibo-fnd-acc-cur;hasPrice"/>
+		<rdfs:label>has minimum redemption price</rdfs:label>
+		<rdfs:domain rdf:resource="&fibo-sec-eq-eq;EquityRedemptionProvision"/>
+		<rdfs:range rdf:resource="&fibo-fbc-fi-ip;MarketPrice"/>
```
rivettp, line 954: As per earlier comment MarketPrice is wrong here

SEC/Equities/EquityInstruments.rdf
```diff
@@ -537,6 +612,26 @@
 		<skos:definition>right that give the shareholder higher priority with respect to certain pre-defined dividend payments over other shareholders and that pay out first in the case of liquidation</skos:definition>
 	</owl:Class>
 	
+	<owl:Class rdf:about="&fibo-sec-eq-eq;PreferredDividend">
+		<rdfs:subClassOf rdf:resource="&fibo-sec-eq-eq;Dividend"/>
+		<rdfs:subClassOf>
+			<owl:Restriction>
+				<owl:onProperty rdf:resource="&fibo-sec-eq-eq;hasFixedDividendRate"/>
```
rivettp, line 619: But a preferred dividend might be variable surely.

rivettp:
(I didn't get a chance to add review comments)
I think it's worth making minor changes to such as use of MarketPrice (in a couple of places) and the DividendSchedule.

SEC/Equities/EquityInstruments.rdf
```diff
@@ -273,7 +287,7 @@
 		<rdfs:subClassOf>
 			<owl:Restriction>
 				<owl:onProperty rdf:resource="&fibo-fnd-dt-fd;hasSchedule"/>
-				<owl:onClass rdf:resource="&fibo-sec-eq-eq;DividendOrShareCouponSchedule"/>
+				<owl:onClass rdf:resource="&fibo-sec-eq-eq;DividendSchedule"/>
```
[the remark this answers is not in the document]
ElisaKendall: They are both optional restrictions.  The declaration is always a dollar amount per share, not a percentage, and there may be a schedule or not, for example, for quarterly payments at that dollar per share amount for some window of time before the board elects to change the amount.  They may have internally agreed on a percentage at the board level, but what is published is typically an amount not a percentage.

SEC/Equities/EquityInstruments.rdf
```diff
@@ -288,11 +302,11 @@
...
 	</owl:Class>
 	
-	<owl:Class rdf:about="&fibo-sec-eq-eq;DividendOrShareCouponSchedule">
+	<owl:Class rdf:about="&fibo-sec-eq-eq;DividendSchedule">
 		<rdfs:subClassOf rdf:resource="&fibo-fnd-pas-psch;PaymentSchedule"/>
 		<rdfs:subClassOf rdf:resource="&fibo-sec-sec-sch;ParametricSchedule"/>
-		<rdfs:label>dividend or share coupon schedule</rdfs:label>
-		<skos:definition>payment schedule indicating the dates on which dividends or share coupon payments are due to be paid</skos:definition>
+		<rdfs:label>dividend schedule</rdfs:label>
+		<skos:definition>payment schedule indicating the dates on which dividends are due to be paid</skos:definition>
```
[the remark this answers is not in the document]
ElisaKendall: The dividend schedule includes the dates on which the dividends are to be paid, not the dates the board decides what to do.  Typically, for common stock, dividends are declared quarterly and there is no schedule per se.  But for preferred shares that have either fixed or adjustable dividends, there may be a regular payment schedule.  The board would still have to authorize the payment, but that's doesn't negate the fact that there is a set payment schedule.

SEC/Equities/EquityInstruments.rdf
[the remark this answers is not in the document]
ElisaKendall: Per discussion in the FCT, and confirming with Citi, a given dividend, especially for a common share, is specified as an amount, and if they need to, they compute a percentage.  But boards of directors typically approve a dollar amount.  The optional schedule is more commonly associated with a preferred dividend, though some boards approve quarterly dividends for the coming year for common shares.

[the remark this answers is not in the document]
ElisaKendall: For preferred shares, it might be stated as a percentage of the par value, but for common stock it is a dollar amount (confirmed with Citi).  But if you state it one way, you can derive the other.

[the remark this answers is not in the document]
ElisaKendall: As a consequence of the discussion, we've made the dividend amount (monetary amount) required for ordinary dividends, but have not added that restriction to preferred dividends, where it may be expressed as a percentage or as an amount.

SEC/Equities/EquityInstruments.rdf
ElisaKendall: We changed it to monetary price to simplify, although in order to redeem the share it is likely required to have evidence from the market, possibly with more than one pricing source.

SEC/Equities/EquityInstruments.rdf
ElisaKendall: One can derive a percentage from an amount, or vice versa, but it could be declared either way.  Most of the definitions we found were for an amount over par rather than a percentage.  In either case you would need to know the par value in order to calculate a percentage, which is not necessarily required if you have a straight amount.  Redemption provisions typically applies to preferred or restricted shares, not to a  common share.

SEC/Equities/EquityInstruments.rdf
ElisaKendall: Earliest date is typically a date - it would be helpful if you can explain why you think it would be relative.  That's not what we are seeing in the literature.  And, it's optional, so if there is some sort of rare relative date trigger, someone could add that to the ontology in a specific case using the hasRelativeDate property.

SEC/Equities/EquityInstruments.rdf
ElisaKendall: This is just to represent a straight rate.  If one would need to represent a formula, you could use hasExpression on the adjustable rate dividend class, but this property wouldn't be the way to do that.  You would do that on the class,  which one can do as needed.  Do you have counter examples that would show why this additional property is needed?

SEC/Equities/EquityInstruments.rdf
ElisaKendall: Yes - restricted common shares.

SEC/Equities/EquityInstruments.rdf
ElisaKendall: Probably not, but how important is it to exclude the case where they are both false?

SEC/Equities/EquityInstruments.rdf
ElisaKendall: For 'who can extend', sometimes an issuer will extend it because of anticipated corporate actions.

For other dates with respect to redemption, sometimes an issuer will specify windows within which one can redeem something so that it only happens when convenient, not on an ad hoc basis whenever.

SEC/Equities/EquityInstruments.rdf
ElisaKendall: The premium, if there is one, is specified as a provision of the contract.  Interest doesn't apply, and for preferred shares, the dividend is calculated based on the nominal value at issuance, not the cumulative face value of the share.

SEC/Equities/EquityInstruments.rdf
ElisaKendall: Those properties are on the adjustable rate dividend itself, rather than on the share.

SEC/Equities/EquityInstruments.rdf
ElisaKendall: Yes, several pairs are disjoint, for example cumulative and non-cumulative, participating and non-participating, etc.

SEC/Equities/EquityInstruments.rdf
ElisaKendall: The property and definition needs to start with a verb, per our policy, especially in the case of an object property.  Please suggest an alternative verb if you don't think 'indicates' is appropriate.

SEC/Equities/EquityInstruments.rdf
ElisaKendall: No - most contracts that have earliest redemption dates specify a precise date.  If you have a reference for your comment it would be great to get the counter example - the FCT did not find one, and this corresponds to what Golden Source and Citi have in their models.

SEC/Equities/EquityInstruments.rdf
ElisaKendall: Ok.  While we think market price is appropriate for some cases, we've loosened it to monetary price.  There is likely a requirement for the more constrained market price from a provenance perspective for evidentiary purposes, but it's possible that a given contract won't require that.

SEC/Equities/EquityInstruments.rdf
ElisaKendall: Most of the time it is fixed rate.  This restriction is optional, however, and there are two subclasses - fixed and adjustable.

SEC/Equities/EquityInstruments.rdf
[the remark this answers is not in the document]
rivettp: (Really need an example)
What is the Commitment then? If in 2019 a Board declared quarterly dividends of $1, $2, none and $1 then are there 4 Dividend instances? Are they created only once a Board has made its determination that quarter? Is there a Schedule at all? How do we capture generally that they are committed to announce a Dividend quarterly (even if it is zero)?

SEC/Equities/EquityInstruments.rdf
rivettp: I read "Normally, there is a waiting period, often five years, between the stock issue date and the first call date". However I now realize that each Preferred Share has a specific issue date so hence a specific early call date.

SEC/Equities/EquityInstruments.rdf
rivettp: But this is *adjustable* rate so needs reference to some sort of formula by which it's adjusted. the explanatoryNote even says "is set by a predetermined formula "

SEC/Equities/EquityInstruments.rdf
rivettp: OK

SEC/Equities/EquityInstruments.rdf
rivettp: Not hugely important but it would seem a neater model to have a en enumeration "RedemptionInitiation" or similar. I'm not generally fan of booleans, since it makes it harder to document the meaning of each case.

rivettp: BTW if it's at ShareholderOption is that the same as a RetractablePreferredShare https://pocketsense.com/redeemable-vs-retractable-preferred-shares-3877.html ?
We should have the synonym.

SEC/Equities/EquityInstruments.rdf
rivettp: I could not google anything for "extendable preferred share": I literally got zero results! Is it really a thing?

SEC/Equities/EquityInstruments.rdf
rivettp: Did you just add them?> My earlier comment was that I could not see such properties.

SEC/Equities/EquityInstruments.rdf
rivettp: OK

SEC/Equities/EquityInstruments.rdf
rivettp: I thought the key point was that the definition can be substituted into a sentence in place of the original name. Let's say we have:
 "XincPrefA hasDistributionMethod Reinvestment"
Using the current definition gives:
 "XincPrefA indicates the method by which dividend payments are to be distributed Reinvestment" 
which does not work at all IMO
So better would be:
 "XincPrefA has method by which dividend payments are to be distributed of Reinvestment"

trypuz (approved):
Approved

SEC/Equities/EquityInstruments.rdf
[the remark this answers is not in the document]
rivettp: Raised issue

SEC/Equities/EquityInstruments.rdf
rivettp: Raised issue #1136

SEC/Equities/EquityInstruments.rdf
rivettp: Raised issue #1137

SEC/Equities/EquityInstruments.rdf
rivettp: Raised #1138

SEC/Equities/EquityInstruments.rdf
rivettp: Raised #1138

Issue #1127, opened 2020-08-19
Extend hygiene tests - avoid ambiguous/vague definitions


rivettp:
In general the presence of " or " in a definition should be treated as a red flag and spur a Warning: a FIBO element should clearly define exactly one concept: the presence of an "or" signals vagueness or optionality.

trypuz:
Agree, @mereolog, could you propose a test for this, please?

mereolog:
This should do the trick:
```
prefix skos: <http://www.w3.org/2004/02/skos/core#>


##
# banner Definitions should not contain the "or" connective.

SELECT ?error ?o
WHERE {
  ?s skos:definition ?o
  FILTER (REGEX(str(?s), "edmcouncil"))
  FILTER (REGEX (str(?o), "\\sor\\s"))
  BIND (concat ("WARN: definition of ", str(?s), " is ambiguous because of the presence of disjunction. ") AS ?error) 
 } 
```

PR #1151, opened 2020-09-14
Hygiene check for disjunctive definitions

mereolog:
Signed-off-by: Pawel Garbacz <pawel.garbacz@makolab.com>

## Description

This is to warn about all definitions that contain the "or" connective.

Fixes: #1127 

## Checklist:

- [X] I'm familiar with the [FIBO developer quide](../CONTRIBUTING.md#contributing-to-the-fibo-code). My contribution meets all the requirements described there.
- [X] My contribution follows the [principles of best practices for FIBO](../ONTOLOGY_GUIDE.md).
- [X] My changes have been reconciled with latest master and no merge conflicts remain.
- [X] This PR is related to exactly one issue. The issue is referenced by using a GitHub keyword such as "fixes", "closes", or "resolves".
- [X] Hygiene tests have been applied by a PR with "(WIP)" in title.
- [X] The issue has been tested locally using a reasoner (for ontology changes).

### DIFF
--- a/etc/testing/hygiene/testHygiene1127.sparql
+++ b/etc/testing/hygiene/testHygiene1127.sparql
@@ -1,0 +1,14 @@
+prefix skos: <http://www.w3.org/2004/02/skos/core#>
+
+
+##
+# banner Definitions should not contain the "or" connective.
+
+SELECT ?error ?o
+WHERE 
+{
+  ?s skos:definition ?o
+  FILTER (REGEX(str(?s), "edmcouncil"))
+  FILTER (REGEX (str(?o), "\\sor\\s"))
+  BIND (concat ("WARN: definition of ", str(?s), " is ambiguous because of the presence of disjunction. ") AS ?error) 
+} 


ElisaKendall:
I think this is a problem, and will fail under a number of valid circumstances.  For example, some credit events apply to a 'party or reference obligation'.   We should not have to duplicate concepts simply because they can apply to more than one concept.  There -may- be cases where the use of 'or' in the definition is an issue, but that's usage specific and can't be applied broadly.

ElisaKendall:
Another situation where 'or' is applicable - for definitions of boolean valued properties, we often say 'whether or not', and that's a reasonable thing to say in the context of a boolean.

rivettp:
The phrase "party or reference obligation" is a good example of where rewording would be valuable - since I have little idea what precisely it means (BTW it does not appear in current FIBO). If it's a restriction I would expect it to be represented via owl:unionOf rather than in the definition text.

In "whether or not" the "or not" is mostly* just noise: OK for conversational talk but totally redundant as far as a definition is concerned. Can anyone think of a case where it adds meaning? * in a minority of cases it could grammatically be replaced by "regardless of whether".

Yes this test (only a warning) could require a bit more care in wording of definitions but I think the benefits of excising imprecision make it worthwhile.

rivettp (approved):

ElisaKendall:
There are many, many cases where the word 'or' is totally appropriate, and I for one am not going to prioritize replacing it everywhere it occurs.  ISO 704 compliance is far more important, as is eliminating circular definitions.  I already have to download the hygiene results into a file to make it searchable every time I open an ontology and start to fix it - this hygiene test will totally overload anything I could possibly do to move things along.

I think it should be handled in a separate file at a minimum, not added to the existing set of hygiene tests, so that it doesn't overburden the work I've been faithfully attempting to do.  But, I want to let everyone know that I OBJECT to this as a test.

Issue #1169, opened 2020-09-22
The ontology guide needs to be updated with respect to why not to use PROV


ElisaKendall:
Some of the issues related to PROV have been addressed, so the language around why we don't use it is old.  This should be clarified so that people don't have the wrong impression.

PR #1170, opened 2020-09-22
OG-1169 - The ontology guide needs to be updated with respect to why not to use PROV

ElisaKendall:

## Description

Updated text related to PROV-O issues in the ontology guide to discuss the issue of punning rather than logical inconsistency

Fixes: #1169 

## Checklist:


rivettp (changes requested):
We should link to the list of OWL-sanctioned datatypes, and an explanation as to why the use of others causes an ontology to be "not logically consistent". BTW is there a difference between that and "cannot be proved to be logically consistent"? In other words are we asserting the negative by being unable to prove the positive (negation by failure)?

Regarding PROV-O we should link to something that explains why punning is bad generally. 
And why PROV-O properties are not reusable (I have reused them outside the FIBO context s am personally interested here)?
[Linking to the issues raised would suffice).

Finally, and again more generally, we should discuss whether these issues could be skirted by use of soft rather than hard reuse of these ontologies, and the pros and cons of these. Discussion here https://papers.ssrn.com/sol3/papers.cfm?abstract_id=3315457
For the purpose of FIBO, where a lot of the usage is for data integration, soft reuse may have a lot of benefits in certain enterprise integration and LOD scenarios. Given that many of these external ontologies are largely structural, and have little to no logical semantics of their own, hard reuse (owl:imports) would seem to be counterproductive.

ElisaKendall:
Pete, do you have time to take a crack at this?  I'm happy to augment / edit what you do but barely had time to fix what was there.  Too many other things on my plate this week.

trypuz (approved):
Approved

rivettp:
@ElisaKendall I should be able to have a go at this next week, but could use a chat with you first e.g. so I can understand the reuse and punning issues you see with PROVO and the general approach of soft reuse I mentioned in my comment (and a topic of the paper I linked to).

ElisaKendall:
@rivettp - Can we agree to let the simple change I made go this week, as there have been folks asking about Prov, and MB reached out to me?  Then we can add content next week or over the next couple of weeks together to improve on that?  I'd be happy to work with you to add that, just can't do that until sometime next week.  Let me know times that could work for you and we can calendar something.

rivettp:
@ElisaKendall what your proposed change says about PROV is "The PROV-O ontology uses punning, and many of its properties are not reusable. Although we have raised issues with the W3C about this"
As someone who has recently used PROV (outside of FIBO) I naturally have concerns about approving this, which will cause concern in my other project. At least without also including justifying detail I can discuss with my project.
- What specifically is the punning? I believe one issue may have been fixed https://lists.w3.org/Archives/Public/public-prov-comments/2017May/0004.html 
- Which properties are not reusable and why? 
- I know you raised issue(s) on OWL Time but what did you raise on PROV? Do you have links?

Regardless of those points, I find it hard to think of an area/use case of FIBO where PROV *would* be useful. So it seems to me the question could indicate a misunderstanding of what either FIBO or PROV is trying to achieve.
If you want to address concerns about the earlier statement on PROV, why not just remove mention of it? 

More generally I don't think it's the place of the Ontology Guide to answer specific such questions. 
We could I guess have a FAQ or suggest people raise issues if they want to propose use of another ontology.

There are many other ontologies which are potential candidates including GLEIF (now), FIGI, W3 Organization, Vcard, QUDV, BPMN, FOAF, Geographical. How much non-reuse do we need to respond to? 
As it happens we've reused nothing we didn't ourselves have a hand in creating, so we could just have a blanket response "We have a low level of trust in others so the principle is Not Invented Here".
And I personally have an aversion to baking in dependencies: we have enough already in FIBO.

We should also describe, and show worked examples, of a mapping approach, to allow integration of other ontologies without hardcoding them into FIBO and creating yet another dependency. I already created a very lightweight FIGI mapping I know you've seen. And we have also discussed doing the same with GLEIF: which would be a new ontology with triples linking the ontology elements via equivalentClass/subClass (properties too).

Bottom line, if you want an immediate change I suggest removing specific mention of PROV: there is IMO no compelling reason to mention it alone, and I cannot even see a case for it even if it had no problems at all.

mereolog:
I've just checked consistency of http://www.w3.org/2006/time#2016 using Hermit 1.4.3.456 and Pellet from within Protege 5.5.0 and found no errors nor inconsistency. So even if it uses several "non OWL-sanctioned types", some people may find it confusing when they will see our inconsistency claim.
IMO use of non OWL-sanctioned types decrease reusability, so perhaps we can give our story just in terms of the latter.

ONTOLOGY_GUIDE.md
```diff
@@ -262,7 +262,7 @@ FIBO is designed as a W3C OWL 2 DL compliant family of ontologies.
...
 
 4. **FIBO is intended to be mapped and reused with other linked data and ontologies.  Having said this, because it is also an international standard, we will not import or reuse any externally developed ontology that is not itself a standard and/or that is not logically consistent**.  
-	* Many people have asked why we don't reuse the **OWL Time ontology from the W3C**, and why we don't use the PROV-O ontology for provenance.  Neither of these ontologies is logically consistent, and although we have raised issues with the W3C about this, their developers have elected not to address these concerns over other needs of their communities.  
+	* Many people have asked why we don't reuse the **OWL Time ontology from the W3C**, and why we don't use the PROV-O ontology for provenance.  OWL Time is not logically consistent due to the use of non-OWL sanctioned datatypes.  The PROV-O ontology uses punning, and many of its properties are not reusable. Although we have raised issues with the W3C about this, their developers have elected not to address these concerns over other needs of their communities.
```
mereolog, line 265: Can we have something like 'OWL Time uses non-OWL sanctioned datatypes, which may limit its reusability.'

mereolog (changes requested):

ONTOLOGY_GUIDE.md
rivettp: @mereolog see my earlier comment regarding PROV-O - I believe a simple fix has been approved for the punning issue (which was very minor and involved 2 properties erroneously typed as AnnotationProperty). I also had some questions for @ElisaKendall regarding the lack of reusability and which issues "we" had raised on PROV-O.
My larger point though (and response to the question "why does FIBO not use PROV-O?) was how it would even make sense in FIBO to use PROV-O? For provenance of what? Can you think of any examples/scenarios?

My other larger point was whether we even need to respond to such questions, and in the Ontology Guide of all places? If someone really thinks another ontology should be used they should raise an issue with a specific proposal and (I suggest after discussion, to save wasted effort) a PR. 

Looking at point 4 as a whole, I would even question the first statement "FIBO is intended to be mapped and reused" since we, nor others AFAIK, have ever attempted that seriously with anything we didn't ourselves develop, as I mentioned in my earlier comment. And I have encountered rejection when proposing that FIBO be published in Linked Open Vocabularies or Linked Open Data Cloud. We are in danger of valuing purity over ubiquity. Maybe we should publish different products called "pure FIBO" and "linkable FIBO" and let the consumer choose, just as we also provide FIBO as (extended)SKOS?

ElisaKendall: @mereolog I'm ok with your change to the sentence about OWL TIme.  And, if you want this in the current release, just delete the comment about PROV, but several people have asked that question specifically, including JPMC, so I answered it.  And, there are other groups reusing FIBO, including the OMG's retail domain task force and Mayo Clinic is considering it as well for some aspects of what they need, so I would not say that we are the only people extending or reusing it.  But, they are mostly reusing FND, BE and FBC, and not other parts of FIBO.

ElisaKendall: Within finance, I can name several banks that are using it now, and I think TradeWeb was working on a PoC that would reuse it.  I don't know who else is doing so - clearly someone should have access to the number of downloads and by whom from the spec.edmcouncil.org/fibo site, though.  It's harder to know statistics from github.

mereolog (approved):
I will the changes suggested in the discussion with @ElisaKendall  and @rivettp in the subsequent PR.

Issue #1319, opened 2021-01-24
An exhaustive set of classes, corresponding to the various equity feature combinations expressed in the CFI standard are needed


ElisaKendall:
Currently we have two of these classes in the main equities ontology, but others are missing and they should all be together in the Equity CFI ontology, corresponding to the things that the codes can classify, per the SEC FCT.

PR #1323, opened 2021-02-01
SEC-73 - An exhaustive set of classes, corresponding to the various equity feature combinations expressed in the CFI standard are needed

ElisaKendall:

## Description

Augmented the reference Equity CFI Classification ontology to include a complete set of classes covering sections 6.2.2 Equities - Common/ordinary shares and 6.2.3 Equities - Preferred/preference shares, for those that are classified as registered securities with respect to the fourth attribute.  This does not yet include all of the codes for preferred shares, only those that apply to common shares.  We anticipate building out those codes as well as the classes covering sections 6.2.4 Equities - Common/ordinary convertible shares, and 6.2.5 Equities - Preferred/preference convertible shares over the coming months.

Note that this change adds 196 classes for preferred shares alone, and, due to numerous hasValue restrictions, increases the time it takes for a DL reasoner to complete processing accordingly.

Fixes: #1319 / SEC-73

## Checklist:

### DIFF
--- a/FBC/DebtAndEquities/Debt.rdf
+++ b/FBC/DebtAndEquities/Debt.rdf
@@ -61,9 +61,9 @@
 		<rdfs:label>Debt Ontology</rdfs:label>
 		<dct:abstract>This ontology defines concepts that are common to all debt instruments, such as debt, borrower, lender, debtor, creditor, interest, principal, and the like. It is designed to be used by various other FIBO specifications, including but not limited to SEC/Debt and LOAN.</dct:abstract>
 		<dct:license rdf:datatype="&xsd;anyURI">http://opensource.org/licenses/MIT</dct:license>
-		<sm:contentLanguage rdf:datatype="&xsd;anyURI">http://www.w3.org/standards/techs/owl#w3c_all</sm:contentLanguage>
-		<sm:copyright>Copyright (c) 2016-2020 EDM Council, Inc.</sm:copyright>
-		<sm:copyright>Copyright (c) 2016-2020 Object Management Group, Inc.</sm:copyright>
+		<sm:contentLanguage rdf:datatype="&xsd;anyURI">https://www.w3.org/TR/owl2-quick-reference/</sm:contentLanguage>
+		<sm:copyright>Copyright (c) 2016-2021 EDM Council, Inc.</sm:copyright>
+		<sm:copyright>Copyright (c) 2016-2021 Object Management Group, Inc.</sm:copyright>
 		<sm:dependsOn rdf:resource="https://spec.edmcouncil.org/fibo/ontology/BE/"/>
 		<sm:dependsOn rdf:resource="https://spec.edmcouncil.org/fibo/ontology/FBC/FinancialInstruments/FinancialInstruments/"/>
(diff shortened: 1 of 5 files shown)


jfgemski (approved):

trypuz:
@ElisaKendall, what is the reason for having prefixes like the "fibo-sec-eq-10962" one?

ElisaKendall:
@trypuz We have a couple of cases that use the ISO code as part of their prefix, the main reason being emphasis that the reference data ontology is specific to that ISO standard.  The EquityCFIClassificationIndividuals ontology, which has this prefix, is intended to cover the equity section of the ISO 10962 CFI standard.  We don't have all of it yet, even for equities, and may not cover codes that reflect certain forms (leaving that to anyone that reuses the ontology and needs those features), but the most common combinations and codes for classification of financial instruments are what we plan to include.  Currently we cover sections 6.2.2, Common/ordinary shares, and for that part of the scheme we have both the classes and codes now, and then for 6.2.3 Preferred/preference shares, this resolution provides an exhaustive set of classes, but not the codes yet.  The intent is to add the codes as well as coverage for 3 more parts of the standard - 6.2.4 Convertible common shares, 6.2.5 Convertible preferred shares, and 6.2.6 Limited partnership units, as well as all of the codes covering each of those parts of the standard.  We consider this to be reference data, specifically supporting the ISO 10962 codes, similar to what we have done for the ISO 4217 currency codes, for example, so that's why the ISO number is part of the prefix.  A number of banks have asked for this over the last few years, so using this prefix makes very clear what the ontology covers.  

We will do the same thing under SEC/Debt eventually (for the classes and codes corresponding to the ISO CFI standard), and for various derivatives areas, as many banks use CFI to help understand the nature of a given instrument.  We didn't do that for the MIC codes - which are in FBC, but that standard splits into two pieces, the business centers and MIC codes themselves, for which we have two ontologies (one for each component).  I think we determined that the business centers in particular might expand beyond what used by the ISO MIC codes standard, and didn't want to limit the scope to the ISO standard, but we should think about whether or not to change the prefix for the MIC codes to fibo-fbc-fct-10383 from the current fibo-fbc-fct-mkti, again for emphasis that the ontology is reference data and covers only the content of the ISO standard.  I can go either way on that one, but the split to two ontologies partly drove the approach there.  Another approach for CFI would be to do something like fibo-fbc-fct-ISO10383, to be even more explicit, but we simply used the ISO standard number in the prefix to be more concise.

ElisaKendall:
@trypuz  One thing this brings up is whether or not our ontology architecture should have some sort of split - 
fibo/ontology/reference, fibo/ontology/example, and fibo/ontology/? maybe just as we have it now, to make even clearer what is reference data and what are examples, although the ontology IRI does make an attempt.  We haven't had a forum for this sort of architectural discussion for quite awhile, and so for now everything is combined under the various domain areas, though we've attempted to include 'example' in the ontology names as appropriate.  If I was starting over, I would likely make the split, but would need everyone to agree before we make a change like that.

tahoeblue (approved):

trypuz:
@ElisaKendall thank you for the explanation. @tahoeblue thank you for your review; Jeff, I added you to the FIBO Write Team; welcome aboard!

trypuz (approved):
Approved

Issue #1344, opened 2021-02-19
Need a property in equities representing the authorized number of shares


ElisaKendall:
We have a number of properties such as shares issued, outstanding, etc. but not the number authorized.

PR #1345, opened 2021-02-19
SEC-149 - Need a property in equities representing the authorized number of shares

ElisaKendall:

## Description

Added a property for specifying the number of authorized shares and restriction on equity issuer saying that they have to have authorized at least some number of shares to be an issuer; adjusted the definition of has shares issued to eliminate confusion between that property and shares authorized.

Fixes: #1344 / SEC-149

## Checklist:

### DIFF
--- a/SEC/Equities/EquityInstruments.rdf
+++ b/SEC/Equities/EquityInstruments.rdf
@@ -124,13 +124,13 @@
 		<owl:imports rdf:resource="https://spec.edmcouncil.org/fibo/ontology/SEC/Securities/SecuritiesRestrictions/"/>
 		<owl:imports rdf:resource="https://www.omg.org/spec/LCC/Countries/CountryRepresentation/"/>
 		<owl:imports rdf:resource="https://www.omg.org/spec/LCC/Languages/LanguageRepresentation/"/>
-		<owl:versionIRI rdf:resource="https://spec.edmcouncil.org/fibo/ontology/SEC/20210101/Equities/EquityInstruments/"/>
+		<owl:versionIRI rdf:resource="https://spec.edmcouncil.org/fibo/ontology/SEC/20210201/Equities/EquityInstruments/"/>
 		<skos:changeNote>The https://spec.edmcouncil.org/fibo/ontology/SEC/20180801/Equities/EquityInstruments.rdf version of this ontology was revised to replace &apos;publicly-traded share&apos; with &apos;exchange-specific share&apos;, which is the more commonly used designation and corresponds better with the intended semantics of this concept, to merge in concepts that were formerly in a separate ShareTerms ontology, and eliminate deprecated elements.</skos:changeNote>
 		<skos:changeNote>The https://spec.edmcouncil.org/fibo/ontology/SEC/20190901/Equities/EquityInstruments.rdf version of this ontology was revised to refine the definition of listed share, update definitions to remove leading articles, add missing properties and restrictions, revise the definition of dividend.</skos:changeNote>
 		<skos:changeNote>The https://spec.edmcouncil.org/fibo/ontology/SEC/20191201/Equities/EquityInstruments.rdf version of this ontology was revised to refine the definition of share to include a restriction for hasSharesOutstanding, eliminate duplication of concepts in LCC, and add the concept of an equity issuer.</skos:changeNote>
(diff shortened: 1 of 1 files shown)


rivettp (approved):

SEC/Equities/EquityInstruments.rdf
```diff
@@ -359,6 +359,12 @@
 				<owl:someValuesFrom rdf:resource="&fibo-sec-eq-eq;Share"/>
 			</owl:Restriction>
 		</rdfs:subClassOf>
+		<rdfs:subClassOf>
+			<owl:Restriction>
+				<owl:onProperty rdf:resource="&fibo-sec-eq-eq;hasSharesAuthorized"/>
+				<owl:someValuesFrom rdf:resource="&xsd;nonNegativeInteger"/>
+			</owl:Restriction>
+		</rdfs:subClassOf>
 		<rdfs:label>equity issuer</rdfs:label>
```
jfgemski, line 368: I don't know why we need equity issuer.  Implies we'll also have debt issuer, derivative issuer, etc.  I believe we only need a class for "issuer".  From Investopedia:
An issuer is a legal entity that develops, registers and sells securities to finance its operations. Issuers may be corporations, investment trusts, or domestic or foreign governments. Issuers are legally responsible for the obligations of the issue and for reporting financial conditions, material developments and any other operational activities as required by the regulations of their jurisdictions.

tahoeblue:
John’s observation sounds reasonable to me.

ElisaKendall:
@tahoeblue @jfgemski Issuer is a role that a legal entity plays, so we have that connection in the model (and have had for ages), and in the case of an issuer that issues equities, they have to have the legal basis for doing so, including some number of authorized shares, etc.  Your definition of legal entity is quite close to what we have, but the pattern we are using is that a legal entity may play many roles - employer, supplier, tenant, borrower, account holder, and issuer all at the same time, for example, and there are properties associated with each of those roles.  A legal entity that is a stock corporation can be an equity issuer, but a legal entity that is an LLC cannot.  So this notion of equity issuer is part of that lattice we're building up in a richer graph.

rivettp:
Well an LLC can issue equity, just not in the form of stock https://carta.com/blog/what-is-llc/#:~:text=Rather%20than%20issuing%20stock%20options,terms%20of%20your%20operating%20agreement.
If the point is to capture such rules, which I think are jurisdiction-specific, we don't seem to be attempting that. Even if we did, it would not need a class for StockIssuer, you could have a restriction on Issuer based on the class of what's issued.
And we would not need a separate individual for the role.

ElisaKendall:
@rivettp To address this point, we opened SEC-151 today in JIRA to rename EquityIssuer to ShareIssuer (agreed on the SEC FCT working group session).

tahoeblue:
SecurityIssuer perhaps?  (Bonds don’t normally have “shares” , for example)

Or, was this perhaps not a promotion of “issuer” to a higher level in the ontology, and just a rename in the equity section?

ElisaKendall:
@tahoeblue We already have Issuer at the higher level, which we use in Bonds.  This is just for the properties related to the number of shares a share issuer has on their books - authorized, issued, outstanding and treasury.

tahoeblue:
Understood, thanks.  This relates more to John Gemski’s suggestion that a subclass of issuer wasn’t needed at the equity level.  Wouldn’t the type of share classes or number of shares issued still relate to “Issuer” and security type ?

ElisaKendall:
@tahoeblue Yes, with respect to the number of shares issued, and they do.  But those same properties do not necessarily apply to  bond issuers or derivatives issuers, for which we use the higher level issuer class.  For example, the issuer of a muni bond doesn't typically issue shares because they are a government entity.  So for that we use the higher level issuer class unless there are properties specific to being a muni bond issuer (there are not at the moment).  Remember that we have not flattened the hierarchy in the way that CFI or other models do - we use roles to connect all kinds of dots, and having a subclass of equity issuer (which we will rename to share issuer in another github/JIRA issue), doesn't mean that we need to exhaustively expand the various kinds of issuers to cover each kind of security.  We only need to do that if there are specific properties related to an issuer of a specific kind of security, which is the case for shares of stock, but may not be otherwise.

jfgemski (approved):

Issue #1436, opened 2021-03-10
Update the MIC codes per the latest version published by ISO (ISO 10383)


ElisaKendall:
The ISO market identifier codes (MIC) for exchanges are updated monthly. This issue is about updating what we have in FIBO to correspond to the latest monthly update for inclusion in the Q1 2021 release.

PR #1437, opened 2021-03-10
FBC-278 - Update the MIC codes per the latest version published by ISO (ISO 10383)

ElisaKendall:

## Description

Update the business centers and markets reference data to reflect the latest updates to the ISO 10383 MIC codes

Fixes: #1436 / FBC-278

## Checklist:

### DIFF
--- a/FBC/FunctionalEntities/BusinessCentersIndividuals.rdf
+++ b/FBC/FunctionalEntities/BusinessCentersIndividuals.rdf
@@ -44,12 +44,11 @@
 		<dct:abstract>This ontology includes individuals representing the set of international business centers corresponding to those identified in FpML as well as additional municipalities called out in the ISO 10383 Codes for exchanges and market identification (MIC) standard.
 		This set of business centers is current with respect to the FpML published XML data as of 2018-06-29.  Note that we have deviated from the standard FIBO naming convention of strict use of camel case to add underscores in city names for readability purposes.</dct:abstract>
 		<dct:license rdf:datatype="&xsd;anyURI">http://opensource.org/licenses/MIT</dct:license>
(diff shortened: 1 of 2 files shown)


rivettp (approved):

jfgemski:
I disagree with this approach unless EDMC commits to have someone periodically update the list based on recent changes.  The same is true for other official lists such as the country codes.  I'd rather show a few entries as examples and then point to the document where the reader can access the latest updates.
In addition, some of these lists are proprietary.  ISO forbids the sharing of codes and we could be open to copyright infringement if we provide the full list.

ElisaKendall:
@jfgemski  Hi John, I understand your concern, but it is unfounded in this case.  OMG has a liaison relationship with ISO, and for the purposes of the standard we have permission to do this.  We've been publishing these codes for several years now, and ISO TC 68, which is the working group responsible for the codes, knows that we are doing so with permission.  We also have a paid license to the standard to that end.  I'm the back-up representative from OMG to TC 68 as it happens.  Richard Beatch with Bloomberg is primary.

trypuz:
> @jfgemski Hi John, I understand your concern, but it is unfounded in this case. OMG has a liaison relationship with ISO, and for the purposes of the standard we have permission to do this. We've been publishing these codes for several years now, and ISO TC 68, which is the working group responsible for the codes, knows that we are doing so with permission. We also have a paid license to the standard to that end. I'm the back-up representative from OMG to TC 68 as it happens. Richard Beatch with Bloomberg is primary.

@ElisaKendall, I'll contact you via e-mail to learn more about the permission you've mentioned in your comment. @jfgemski  may have a point.

dsnewman (approved):

Issue #1471, opened 2021-03-26
hasNominalCapital should be a subproperty of hasMonetaryAmount in Corporations


ElisaKendall:
The explanatory note on this property also needs to be reworded or removed.  Incorporation and registration dates should also be made explicit dates rather than simple dates.

PR #1472, opened 2021-03-29
BE-233 - hasNominalCapital should be a subproperty of hasMonetaryAmount in Corporations

ElisaKendall:

## Description

Made hasNominalCapital a subproperty of hasMonetaryAmount and cleaned up the explanatory note.  Also made  incorporation and registration dates explicit dates rather than simple dates.

Fixes: #1471 / BE-233

## Checklist:

### DIFF
--- a/BE/Corporations/Corporations.rdf
+++ b/BE/Corporations/Corporations.rdf
@@ -66,7 +66,7 @@
 		<owl:imports rdf:resource="https://spec.edmcouncil.org/fibo/ontology/FND/Relations/Relations/"/>
 		<owl:imports rdf:resource="https://spec.edmcouncil.org/fibo/ontology/FND/Utilities/AnnotationVocabulary/"/>
 		<owl:imports rdf:resource="https://www.omg.org/spec/LCC/Languages/LanguageRepresentation/"/>
-		<owl:versionIRI rdf:resource="https://spec.edmcouncil.org/fibo/ontology/BE/20210101/Corporations/Corporations/"/>
+		<owl:versionIRI rdf:resource="https://spec.edmcouncil.org/fibo/ontology/BE/20210301/Corporations/Corporations/"/>
 		<skos:changeNote>The https://spec.edmcouncil.org/fibo/ontology/BE/20131101/Corporations/Corporations.rdf version of this ontology was modified per the issue resolutions identified in the FIBO BE 1.0 FTF report.</skos:changeNote>
 		<skos:changeNote>The https://spec.edmcouncil.org/fibo/ontology/BE/20160201/Corporations/Corporations.rdf version of this ontology was modified per the FIBO 2.0 RFC.</skos:changeNote>
 		<skos:changeNote>The https://spec.edmcouncil.org/fibo/ontology/BE/20180901/Corporations/Corporations.rdf version of this ontology was modified per the FIBO 2.0 RFC to generalize certain unions where they were no longer required.</skos:changeNote>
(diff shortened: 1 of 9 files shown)


jfgemski (approved):

rivettp (changes requested):
Happy with the ExplicitDate changes but not the primary change - I don't think nominalCapital should be a sub property of hasMonetaryAmount - since the authorized shares don't represent the capital that an org has access to - which is the issued shares.

ElisaKendall:
@rivettp But hasMonetaryAmount is simply saying that the property, hasNominalCapital, is a monetary amount.  It doesn't say that the organization actually has that amount of money available to it, nor does the definition.

trypuz:
Let me remind you that you can see the content of this PR here: 

- https://spec.edmcouncil.org/fibo/ontology?version=pr-1472%2Flatest

E.g.:
- https://spec.edmcouncil.org/fibo/ontology/BE/Corporations/Corporations/hasNominalCapital?version=pr-1472%2Flatest

jfgemski (approved):

rivettp:
The actual definition of hasMonetaryAmount "specifies a value of something" does not match what you wrote @ElisaKendall - the value of an entity is if anything its market cap not the number of authorized shares which is theoretical. We already indeed have hasMarketCapitalizationValue as a sub property of hasMonetaryAmount.

rivettp:
If we merely want to say that hasNominalCapital is a MonetaryAmount we have the range of the property for that.

ElisaKendall:
@rivettp Exactly.  Thus your objection doesn't make sense to me.  The restriction that adds hasNominalCapital to a corporation is min 0.  The fact that a board has authorized some number of shares that themselves have some value, even if miniscule, is what is used to calculate the valuation of a company in pre-IPO stages.  There are no publicly issued shares that can be used for that purpose at that point in the lifespan of the organization.  What is important are the number of authorized shares times the perceived guestimate of the value of the shares (having done this with more attorneys than employees at one point in my career).  Thus, hasNominalCapital is important for that purpose if nothing else and indeed should be a subproperty of hasMonetaryAmount, even if that monetary amount represents a book value rather than actual money available to the organization.

rivettp:
Even pre-IPO the amount of authorized capital is never used for a valuation. Even though there are no publicly-issued shares, the value of privately issued shares is a good indication; as is the proposed number of shares to ISSUE (generally a lot less than the AUTHORIZED number), combined with some sort of multiple of sales/turnover/profit.
My point is that if someone queries hasMonetaryAmount, multiplied by percentage owned, across a set of assets, this hasNominalCapital value will be utterly useless and misleading. Further, it will clutter up the result from the useful hasMarketCapitalization sub property. What is the use case where it would ever be useful to have hasNominalCapital too?

ElisaKendall:
@rivettp You may be thinking of companies that are close to IPO or registering for an IPO, not for start-up organizations in early stages.  The calculations for valuation made by our investors in Sandpiper included all authorized shares and were not limited to those that had been issued.

ElisaKendall:
@rivettp I think the right way to resolve this is to change this property, both its name and range, to hasAuthorizedShares, move the restriction to the corporate bylaws (since that's where it belongs, and since modifying that number requires a board resolution), and then leave it at that, and ignore any potential value associated with it.

rivettp:
I'm happy with your last proposal, though not sure we need to make a change since we already have fibo-sec-eq-eq:hasSharesAuthorized

[I wrote the following prior to seeing your proposa - but keeping the reference in case it's helpful. We may also want par/nominal value per share alongside hasSharesAuthorized]
See https://capbase.com/par-value-fmv-how-to-price-your-startups-common-shares/ which says "Par value, also called "face value" or "nominal value," is the lowest legal price for which a corporation may sell its shares, and is typically set between a fraction of a cent (the lower range is known as a low par value) and a few dollars. This has little to nothing to do with how much a corporation’s shares are actually worth or sold for. "
So again, as per the last sentence above, is not any meaningful "value" even for startups, so should not be included in hasMonetaryAmount since that is about the value.

ElisaKendall:
@rivettp - I could move hasSharesAuthorized to corporations, and add the restriction on bylaws using that property then, if you are ok with that? I think we have hasNominalValue as a property elsewhere.

tahoeblue:
I have not been following this thread from the start, but “market cap” does not typically include authorized shares. Investors may well be interested in authorized shares in order to take into account future dilution should authorized shares be issued or awarded, as well as what amount would be appropriate for a public offering.  But computing a “nominal value” by using a market price of tradable shares and including shares in the product that are not on the market compromises the basis of the market price discovery, and hence the market price, and hence the “nominal value”.

ElisaKendall:
@tahoeblue Thanks - that was timely and helps.  I'm renaming the property that was hasNominalCapital to hasNominalValue, which I thought we had but we do not, and will move it to financial instruments, since our definition of nominal value is about instruments and not other things.  But, I'm also moving hasSharesAuthorized from SEC to CorporateBodies, where we define a corporation, and add a restriction to corporate bylaws to cover shares authorized.  In the U.S. the number of authorized shares is stated in corporate bylaws and can only be modified by board resolution and I think a vote of the existing shareholders.  So it really belongs in corporations (or corporate bodies) not in equities.

rivettp (changes requested):
Happy with all, except moving hasSharesAuthorized to the Executives ontology. Which has abstract "This ontology defines concepts relating to executives and their formal capacities.". Just because they are authorized by the Board does not justify this - since the Board ultimately decides everything. Why not have this property in Corporations where its predecessor and hasIssuedCapital are?

tahoeblue:
Pete’s suggestion sounds right to me.   Also,”executives” are generally part of management, not the Board.

ElisaKendall:
@rivettp @tahoeblue I can argue for either location.  Some of the content in executives, such as the definition of certain incorporation documents including corporate bylaws, is why I put it there, but it's not important enough to argue the point.  More importantly, this property is only used in executives and equity instruments, so I thought executives was a good home for it.  I've now moved it though.

rivettp (approved):

trypuz (approved):
Approved

Issue #1479, opened 2021-04-14
Complete the situational lattice for the 'long way around' the pattern


ElisaKendall:
The situational lattice is missing one set of property chains to go 'the long way around' the lattice to cover independent party to independent party relations through the lattice. This pair of properties is key to refining / completing patterns for ownership and control, and for analysis of AML-based and other linkages.

PR #1480, opened 2021-04-14
FND-341 - Complete the situational lattice for the 'long way around' the pattern

ElisaKendall:

## Description

Augmented the Parties ontology with two additional property chains that complete the 'long way around' the lattice for use in AML and other use cases

Fixes: #1479 / FND-341

## Checklist:

### DIFF
--- a/FND/Parties/Parties.rdf
+++ b/FND/Parties/Parties.rdf
@@ -72,7 +72,7 @@
 		<owl:imports rdf:resource="https://spec.edmcouncil.org/fibo/ontology/FND/Relations/Relations/"/>
 		<owl:imports rdf:resource="https://spec.edmcouncil.org/fibo/ontology/FND/Utilities/AnnotationVocabulary/"/>
 		<owl:imports rdf:resource="https://www.omg.org/spec/LCC/Languages/LanguageRepresentation/"/>
-		<owl:versionIRI rdf:resource="https://spec.edmcouncil.org/fibo/ontology/FND/20210301/Parties/Parties/"/>
+		<owl:versionIRI rdf:resource="https://spec.edmcouncil.org/fibo/ontology/FND/20210401/Parties/Parties/"/>
 		<skos:changeNote>The http://www.omg.org/spec/EDMC-FIBO/FND/20130801/Parties/Parties.rdf version of the ontology was was modified per the issue resolutions identified in the FIBO FND 1.0 FTF report and in https://spec.edmcouncil.org/fibo/ontology/FND/1.0/AboutFND-1.0/.</skos:changeNote>
 		<skos:changeNote>The http://www.omg.org/spec/EDMC-FIBO/FND/20141101/Parties/Parties.rdf version of this ontology was revised as a part of the issue resolutions identified in the FIBO FND 1.1 RTF report to add a parent of hasDate to date properties.</skos:changeNote>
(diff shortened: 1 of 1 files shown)


dallemang (approved):
It's probably worth commenting somewhere that these chains are generalizations of things like "owns" and "is owned by" (assuming that this is correct; that's why we need the example); this will make it clearer what is going on here.

ElisaKendall:
@dallemang You're exactly right - that's what these are for, and this is the start of a series of revisions to better integration our properties, eliminate a few that are unnecessary, etc.  The ownership example will be done in the next round of updates (attempting to do this one pattern/example at a time).  Thanks!

dallemang:
There has always been some contention around whether (for example) owns/isOwnedBy links a relative thing to another relative thing or an independent party to another independent party.  This can be formally modeled in OWL, so we have to be clear about which one it is.   This should be a worked example in the next issue.

tahoeblue:
Can it not be both, or either ?

dallemang:
That's where the contention comes in - logically, it can indeed be either (both doesn't make a lot of sense).   So we need a policy; e.g., we fill in one of each (and then we have to name all of them - ugh!), or we always fill in the relative/relative (respectively independent/independent); or we do either one, and document it (formally) in the model (if that's the case, we still have to do both in the template, which is what Elisa is working on now).   I don't have a strong opinion on which of these policies we want to recommend, but I do want us to be clear about whatever policy we choose.

ElisaKendall:
@dallemang @tahoeblue When I update the ownership pattern, I'll also update the examples to make it clear.  Right now, it is definitely not clear as Dean says above.  I have to draw a bunch of pictures every time I add a new entity and want to include their ownership details to remind myself of how it works, and then re-run the reasoner to make sure it all worked properly as well as run a handful of queries.  And, if it takes me forever to figure it out, the likelihood of someone else being able to do so is quite low.  The pattern needs to be clean and the examples will reflect that once I've fixed it.

dsnewman (approved):

tahoeblue:
I am no doubt exhibiting my ignorance of specific semantic technology definitional details, but I’m not sure why both independent things and relative things can not each have “ownership" relationships.  I am quite open to be schooled on the topic, however ! :)

dallemang:
Your proposal (use the same property for relations between independent parties as between relative ones) is another possibility, and there is no technical barrier to it.  Elisa could indeed use the same property in both contexts.   The downside of this approach is that for someone who is already a bit confused about the difference between an independent and relative thing (which is everyone when they first come to FIBO; it's a subtle distinction), this invites confusion.   

The point of my original comment was that whatever policy we choose, we need to document it clearly and use it consistently.  This PR is the place to make this point, since this activity is where we are designing the template that these things will be mapped to.

rivettp:
But @dallemang we removed classes IndependentThing and RelativeThing from FIBO Foundations since they were too academic and irrelevant; not to mention unhelpful and often leading to inconsistencies (since whether something is relative not infrequently depends on context and time frame).

dallemang:
IndependentParty is still there, and the pattern is still in use.     @ElisaKendall  can correct me if I am wrong, but my understanding of the "situational lattice" is that this is a specific pattern where we are using these things in a particular way.  my original comment was a request to document this clearly and unambiguously, so that we know how this pattern relates to specific instances of it  (like owns and controls) so that discussions like this aren't just abstract; we know how they look with relatiosnhips we  (and other users) are familiar with.

ElisaKendall:
@dallemang That's correct, independent party is still in use, and 'ThingInRole' is essentially 'relative thing', and contract and situation are both 'mediating thing's.  Situation addresses the need @rivettp mentions with respect to context and time, which 'mediating thing' did not really do.

Issue #1483, opened 2021-04-15
Revise the ownership lattice relations to better align with the situational model


ElisaKendall:
Most properties covering ownership were updated and realigned with the original update to situation, but there are still challenges with respect to whether an owner is a party or role in some cases. The pattern should be cleaned up for the sake of consistency and ease of use based on the latest additions to the situation properties.  Some number of examples in FBC may be impacted by any change in property names to address this.

PR #1484, opened 2021-04-16
FND-342 - Revise the ownership lattice relations to better align with the situational model

ElisaKendall:

## Description

This resolution involved renaming certain ownership properties to better align with the situational pattern, to improve consistency and clarify usage, including: 
(1) hasOwnedThing --> hasOwnedAsset (name was inconsistent with target range), 
(2) isOwnedThingInRole --> isOwnedAsset (inverse of hasOwnedAsset), 
(3) isOwningPartyInRole --> isOwningParty (name was inconsistent with its inverse), 
(4) isOwnedBy --> isAssetOf (for consistency with the pattern - points from asset to owner), 
(5) added ownsAsset - inverse of isAssetOf (party in role to party in role, was a gap), 
(6) made owns a subproperty of 'plays active role that directly affects', and 
(7) added a new isOwnedBy, subproperty of experiences directly and inverse of owns to complete the pattern (was also a gap) and create a consistent pattern of naming the various roles involved in ownership.

The resulting pattern is shown in the figure attached:  

[OwnershipPattern.pdf](https://github.com/edmcouncil/fibo/files/6322840/OwnershipPattern.pdf)

See this PR in [FIBO Viewer](https://spec.edmcouncil.org/fibo/ontology/FND/OwnershipAndControl/Ownership/?version=pr-1484%2Flatest)

Fixes: #1483 / FND-342

## Checklist:

### DIFF
--- a/FBC/ProductsAndServices/ClientsAndAccounts.rdf
+++ b/FBC/ProductsAndServices/ClientsAndAccounts.rdf
@@ -110,7 +110,7 @@
 		<owl:imports rdf:resource="https://spec.edmcouncil.org/fibo/ontology/FND/Utilities/AnnotationVocabulary/"/>
 		<owl:imports rdf:resource="https://www.omg.org/spec/LCC/Countries/CountryRepresentation/"/>
 		<owl:imports rdf:resource="https://www.omg.org/spec/LCC/Languages/LanguageRepresentation/"/>
-		<owl:versionIRI rdf:resource="https://spec.edmcouncil.org/fibo/ontology/FBC/20210301/ProductsAndServices/ClientsAndAccounts/"/>
+		<owl:versionIRI rdf:resource="https://spec.edmcouncil.org/fibo/ontology/FBC/20210401/ProductsAndServices/ClientsAndAccounts/"/>
 		<skos:changeNote>The https://spec.edmcouncil.org/fibo/ontology/FBC/20160801/ProductsAndServices/ClientsAndAccounts.rdf version of this ontology was revised per the FIBO 2.0 RFC with respect to the definitions for accounts and account identifiers, such as BBAN and IBAN identifiers, including but not limited to bank accounts.</skos:changeNote>
 		<skos:changeNote>The https://spec.edmcouncil.org/fibo/ontology/FBC/20180801/ProductsAndServices/ClientsAndAccounts/ version of this ontology was modified to support the addition of maturity-related properties to financial instruments.</skos:changeNote>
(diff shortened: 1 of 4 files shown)


dallemang (commented):
In this version (as with previous versions), owner hasIdentity exactly 1 thing, but also owner is a subclass of* actor, and actor hasIdentity exactly 1 Independent party; so the restriction to exactly 1 thing at owner is redundant.  

This issue has been around for a while, but I think it is relevant to this jira, since this restriction is part of the pattern that is being adjusted here. 

I think the right fix is to just remove the restriction hasIdentity exactly 1 Thing, since it is already taken care of up at party in role. 

It occurs to me that there is a more pervasive error here, but that wouldn't be for this jira.  When we say that party in role hasIdentity exactly 1 independent party, OWL allows for a particular party in role to have multiple identities, as long as exactly 1 of them is an independent party, e.g., it could have an identity that is a situation or a commitment or an offering or some other inappropriate thing (and hasIdentity doesn't have any range information to prevent that).    I'm not sure what the ramifications of this are, but it does seem like the model is too permissive in this way.  The fix would be to split such restrictions (once we figure out which ones they are; not every exactly 1 restriction has this problem) into exactly 1 Thing and allvaluesfrom IndependentParty.

ElisaKendall:
@dallemang I don't see that restriction on owner - it inherits the restriction on hasIdentity as you mention, but the only restriction on that class is that it is played by owns something.  Having said that, the restrictions weren't aligned with the pattern, and there is a redundant restriction on Asset with respect to hasIdentity - I've addressed that.  With respect to the latter issue, we might need to do that higher up the food chain - not on owner or asset, but at the party in role level.  I'm not sure how much of an issue that would be, but if we decide to address it I would do so at a higher level.

dallemang:
You're right - I was looking sometimes at the Q1 and sometimes at the PR (I wrote to Robert about how I can get a hold of the PR products that would make it easier for me to review the details in the most recent version).   I'm glad you found it. 

I'm going to wait for the commit to finish building so that I can look at the result in the FIBO Viewer before I approve it.

ElisaKendall:
@dallemang Ok - thanks!

dallemang (approved):
This change fixes not only the issue on this jira, but (at least in this case) the larger issue of qualified cardinality.  The new model has a restrioction  =1 Thing inherited from Thing in Role, and then specializations of that  (=1 Agent and =1 Party) for the subclasses.

dsnewman (approved):

Issue #1494, opened 2021-04-21
Need to sort out the schizophrenic relationship between independent party and person/organization


ElisaKendall:
Currently, person and organization are defined separately from independent party, requiring declaration of various relationships, such as organizational membership in the parties ontology that add subclass relationships, etc. rather than in organizations, where that particular one belongs.  It also complicates our ability to build out other situations and corresponding relationships, and adds to inefficiencies in query formulation and reasoning.

This needs to be fixed, and there are common constructs, such as having a name, address, and maybe others that apply to all independent parties that person and organization could then inherit. But right now this isn't clean and needs to be sorted out.

PR #1495, opened 2021-04-21
FND-343 - Need to sort out the schizophrenic relationship between independent party and person/organization

ElisaKendall:

## Description

Eliminated a number of architectural issues related to parties, organizations, and people by reversing the dependencies between them, and moved organization member from parties to organizations as a consequence; also revised tests in etc to align with the latest versions of these ontologies and eliminated vacuous test ontologies (i.e. those that only had imports statements but no individuals), as they included dependencies on the ontologies modified by this resolution, and addressed a level-crossing issue in ownership made apparent by testing this issue resolution.

Fixes: #1494 / FND-343

## Checklist:


dallemang (approved):
The changes look extensive, but that's really just because changing a name for something that is referenced a lot has widespread impact.

jfgemski (approved):

FND/Organizations/Organizations.rdf
```diff
@@ -103,5 +101,22 @@
...
+				<owl:someValuesFrom>
+					<owl:Restriction>
+						<owl:onProperty rdf:resource="&lcc-lr;isMemberOf"/>
+						<owl:allValuesFrom rdf:resource="&fibo-fnd-org-org;Organization"/>
+					</owl:Restriction>
+				</owl:someValuesFrom>
+			</owl:Restriction>
+		</rdfs:subClassOf>
+		<rdfs:label>organization member</rdfs:label>
+		<skos:definition>identifies an organization or someone that plays a membership role in some organization</skos:definition>
```
rivettp, line 119: No this IS the membership role. The property isPlayedBy identifies who plays the role.

FND/Organizations/Organizations.rdf
```diff
@@ -103,5 +101,22 @@
...
+	
+	<owl:Class rdf:about="&fibo-fnd-org-org;OrganizationMember">
+		<rdfs:subClassOf rdf:resource="&fibo-fnd-pty-pty;PartyInRole"/>
+		<rdfs:subClassOf>
+			<owl:Restriction>
+				<owl:onProperty rdf:resource="&fibo-fnd-pty-rl;isPlayedBy"/>
+				<owl:someValuesFrom>
+					<owl:Restriction>
+						<owl:onProperty rdf:resource="&lcc-lr;isMemberOf"/>
+						<owl:allValuesFrom rdf:resource="&fibo-fnd-org-org;Organization"/>
```
rivettp, line 113: An instance of this class can be a member of multiple organizations then? Is that deliberate? What properties would be associated with the instance? I really don't follow how this would be used in practice - do we have an example?

FND/Organizations/Organizations.rdf
```diff
@@ -86,9 +85,8 @@
 		</rdfs:subClassOf>
 		<rdfs:subClassOf>
 			<owl:Restriction>
-				<owl:onProperty rdf:resource="&fibo-fnd-plc-adr;hasAddress"/>
-				<owl:onClass rdf:resource="&fibo-fnd-plc-adr;PhysicalAddress"/>
-				<owl:minQualifiedCardinality rdf:datatype="&xsd;nonNegativeInteger">0</owl:minQualifiedCardinality>
+				<owl:onProperty rdf:resource="&lcc-lr;hasMember"/>
```
rivettp, line 88: Very unclear how one would use hasPart vs hasMember when representing an org structure/ hierarchy. Especially since these properties are soo generic (from LCC)

FND/OwnershipAndControl/Ownership.rdf
```diff
@@ -126,13 +126,13 @@
 		<rdfs:subClassOf>
 			<owl:Restriction>
 				<owl:onProperty rdf:resource="&fibo-fnd-oac-own;hasOwnedAsset"/>
-				<owl:someValuesFrom rdf:resource="&fibo-fnd-pty-pty;Undergoer"/>
+				<owl:someValuesFrom rdf:resource="&fibo-fnd-oac-own;Asset"/>
 			</owl:Restriction>
 		</rdfs:subClassOf>
 		<rdfs:subClassOf>
 			<owl:Restriction>
 				<owl:onProperty rdf:resource="&fibo-fnd-oac-own;hasOwningParty"/>
```
rivettp, line 134: the name owningParty is misleading since the property is for the role not the party

FND/OwnershipAndControl/Ownership.rdf
```diff
@@ -126,13 +126,13 @@
 		<rdfs:subClassOf>
 			<owl:Restriction>
 				<owl:onProperty rdf:resource="&fibo-fnd-oac-own;hasOwnedAsset"/>
```
rivettp, line 128: this seems to break the rule of including the range in the property name

rivettp (commented):
This is a major change of a very fundamental aspect of Production FIBO. I'm concerned this is getting pushed through with no presentation or time for due consideration, or examples.
In terms of confusion, which this was meant to address, there is still huge inconsistency in the use of "party" to represent the legal entity (as in IndependentParty) and to represent properties linking to a role e.g. hasOwningParty.
Also it's unclear how a class such as OrganizationMember is meant to be used (again no examples) and whether one instance would suffice for someone beign part of many organizations (that is not restricted).

FND/Parties/Parties.rdf
```diff
@@ -113,26 +96,23 @@
...
+		<rdfs:subClassOf>
+			<owl:Restriction>
+				<owl:onProperty rdf:resource="&fibo-fnd-plc-adr;hasAddress"/>
+				<owl:onClass rdf:resource="&fibo-fnd-plc-adr;PhysicalAddress"/>
+				<owl:minQualifiedCardinality rdf:datatype="&xsd;nonNegativeInteger">0</owl:minQualifiedCardinality>
+			</owl:Restriction>
+		</rdfs:subClassOf>
+		<rdfs:label>independent party</rdfs:label>
+		<owl:disjointWith rdf:resource="&fibo-fnd-pty-rl;AgentInRole"/>
+		<skos:definition>any person or organization</skos:definition>
```
rivettp, line 115: Need to say something about the meaning of "independent". Is a baby independent, or a wholly owned subsidiary?

FND/Parties/Parties.rdf
```diff
@@ -113,26 +96,23 @@
...
-		<skos:definition>identifies an organization or someone that plays a membership role in some organization</skos:definition>
+		<rdfs:subClassOf>
+			<owl:Restriction>
+				<owl:onProperty rdf:resource="&fibo-fnd-plc-adr;hasAddress"/>
+				<owl:onClass rdf:resource="&fibo-fnd-plc-adr;PhysicalAddress"/>
+				<owl:minQualifiedCardinality rdf:datatype="&xsd;nonNegativeInteger">0</owl:minQualifiedCardinality>
+			</owl:Restriction>
+		</rdfs:subClassOf>
+		<rdfs:label>independent party</rdfs:label>
+		<owl:disjointWith rdf:resource="&fibo-fnd-pty-rl;AgentInRole"/>
```
rivettp, line 114: No! This utterly screws up legitimate implementations already in production and is a show stopper for me. If you want it for testing the ontology then add it temporarily.

rivettp (commented):
Making AgentInRole disjoint with IndependentParty is a showstopper that unnecessarily prevents useful multiple classification. Why can't Person Pete also be classified as an Owner?

ElisaKendall:
> This is a major change of a very fundamental aspect of Production FIBO. I'm concerned this is getting pushed through with no presentation or time for due consideration, or examples.
> In terms of confusion, which this was meant to address, there is still huge inconsistency in the use of "party" to represent the legal entity (as in IndependentParty) and to represent properties linking to a role e.g. hasOwningParty.
> Also it's unclear how a class such as OrganizationMember is meant to be used (again no examples) and whether one instance would suffice for someone beign part of many organizations (that is not restricted).

Actually, it is not as big a change are you are claiming.  The intent was to untangle the confusion in the class hierarchy per requests from a couple of banks and others including IBM research, who have been working with clients off and on using FIBO. With respect to organization member in particular, we use that class sparingly to date, but it needs to be better integrated with the overall situation pattern, since organization membership is typically something that holds for some period of time, or at least has a defined starting point.  This is a step in that direction, not the entire solution.

FND/Organizations/Organizations.rdf
ElisaKendall: > No this IS the membership role. The property isPlayedBy identifies who plays the role.

Yes, you're right - I didn't want to change everything about the class in this issue, just move it to untangle the circular and reversed dependencies between parties, people, and organizations.

FND/Organizations/Organizations.rdf
ElisaKendall: > An instance of this class can be a member of multiple organizations then? Is that deliberate? What properties would be associated with the instance? I really don't follow how this would be used in practice - do we have an example?

I didn't change that this time around - I think more work is needed here as I said.  I simply moved it from parties to organizations after unwinding the hierarchy.  When we hook it into the pattern properly, I'll create examples.  We do have examples that work for ownership (in FBC, but they exist), but not for organization membership yet.

FND/Organizations/Organizations.rdf
ElisaKendall: > Very unclear how one would use hasPart vs hasMember when representing an org structure/ hierarchy. Especially since these properties are soo generic (from LCC)

hasPart is transitive, but hasMember is not, among other differences.  But you're right - when we work through the organization membership situational pattern we can create a more specific property locally if needed.

FND/OwnershipAndControl/Ownership.rdf
ElisaKendall: > the name owningParty is misleading since the property is for the role not the party

Do you have a suggestion for an alternative?

FND/OwnershipAndControl/Ownership.rdf
ElisaKendall: > this seems to break the rule of including the range in the property name

The other option was hasOwnedThingInRole, which seemed more complicated than it needed to be.  Naming is not easy, and for a complex pattern like this, it's even more difficult to be clear about what you are saying.  If you have a suggestion, please say so.  The current pattern is available in the EDW slides, attached, at the end of the presentation.
[EDW2021-ModelingPatterns-v6.pdf](https://github.com/edmcouncil/fibo/files/6368571/EDW2021-ModelingPatterns-v6.pdf)

BE/OwnershipAndControl/Executives.rdf
```diff
@@ -422,7 +426,7 @@
 	
 	<owl:Class rdf:about="&fibo-be-oac-exec;CorporateOfficer">
 		<rdfs:subClassOf rdf:resource="&fibo-be-oac-exec;Signatory"/>
-		<rdfs:subClassOf rdf:resource="&fibo-fnd-pty-pty;OrganizationMember"/>
+		<rdfs:subClassOf rdf:resource="&fibo-fnd-org-org;OrganizationMember"/>
 		<rdfs:label>corporate officer</rdfs:label>
 		<skos:definition>high-level management executive of a corporation or other organization, hired by the board of directors or the business owner(s), charged with certain operational responsibilities</skos:definition>
```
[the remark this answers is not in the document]
ElisaKendall: > It seems wrong to use the word "executive" in this definition since Executive is a separate class which this does not subtype.

The intent is to look at a number of these relationships and integrate them into the situational pattern appropriately, including adding executive as a parent class.

> Making AgentInRole disjoint with IndependentParty is a showstopper that unnecessarily prevents useful multiple classification. Why can't Person Pete also be classified as an Owner?

This has been in the ontology for at least a couple of years, probably longer.  It has been important for testing data quality for one of my customers, where we have been integrating content from multiple sources to address problems related to identity resolution and other things.  For the use cases we've tested, it works, and it allowed integration with BFO for one government user, which makes a similar distinction.  Not all FIBO users will need to integrate with BFO-based ontologies, of course, but making this distinction is critical if they do.

FND/Parties/Parties.rdf
ElisaKendall: > Need to say something about the meaning of "independent". Is a baby independent, or a wholly owned subsidiary?

That name came from an IBM model years ago - MB used it to appease several FCT members at the time who were also attempting to integrate FIBO with that model.  It's a legacy thing as far as I'm concerned at this point, where 'party' would do, but changing THAT would really have far reaching consequences.

FND/Parties/Parties.rdf
ElisaKendall: > No! This utterly screws up legitimate implementations already in production and is a show stopper for me. If you want it for testing the ontology then add it temporarily.

As I said, it's been there for years, literally.  Most FIBO users don't run reasoners in operational / production settings, only when testing the ontology and looking for issues while it is under development or I'm certain you would have found it ages ago.

Issue #1496, opened 2021-04-23
Eliminate unnecessary properties that are currently on an index


ElisaKendall:
Representation of an index, such as the DJIA, should not include the value of the index itself, which should be representation in a separate value for MarketRate that points to the index.

A ReferenceIndex, in BasketIndices, should be a child of ScopedMeasure rather than of StatisticalMeasure, but we should not include the restriction on hasIndexValue, which should be on the market rate.

MarketRate also needs a more specific property that points to the index - currently we use isAbout from Documents in FND, which could be the superproperty of this new property but is not sufficiently specific. The FCT agreed to use isValueOf for this purpose.

MarketRate does not need the hasPublisher restriction, as it could be a rate quoted at any time with any source, so it should not be a child of PublishedFinancialInformation.

PR #1497, opened 2021-04-23
IND-91 - Eliminate unnecessary properties that are currently on an index

ElisaKendall:

## Description

Eliminated restrictions on reference rate and economic indicator for the value of the corresponding rate or index; added a restriction for isValueOf on the market rate to point to the thing it is a value of; made reference rate a subclass of scoped measure; eliminated the subclass on published financial information from market rate.

Fixes: #1496 / IND-91

## Checklist:

### DIFF
--- a/IND/EconomicIndicators/EconomicIndicators.rdf
+++ b/IND/EconomicIndicators/EconomicIndicators.rdf
@@ -117,7 +117,7 @@
 		<owl:imports rdf:resource="https://spec.edmcouncil.org/fibo/ontology/IND/Indicators/Indicators/"/>
 		<owl:imports rdf:resource="https://www.omg.org/spec/LCC/Countries/CountryRepresentation/"/>
 		<owl:imports rdf:resource="https://www.omg.org/spec/LCC/Languages/LanguageRepresentation/"/>
-		<owl:versionIRI rdf:resource="https://spec.edmcouncil.org/fibo/ontology/IND/20210301/EconomicIndicators/EconomicIndicators/"/>
+		<owl:versionIRI rdf:resource="https://spec.edmcouncil.org/fibo/ontology/IND/20210401/EconomicIndicators/EconomicIndicators/"/>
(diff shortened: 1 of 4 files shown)


rivettp (approved):
Looking at it as a whole I think we need to revisit "market rate" as as a ratio and  "rate that measures market trends for a set of instruments". Or revisit values of DJIA etc as instances of that. Since a single value of DJIA does not indicate a rate (of change) or trend. It;s rather a summary/aggregate of several values in a defined scope used as a proxy/indicator (as in IND!) for that scope. e.g. a value for DJIA is an indicator/proxy of the share prices of large US established companies. A series of such values can indicate a trend - as indeed can a series of closing prices of a single stock - it's the series that gives rise to the trend. BTW DJIA is not technically an "average" even since the share prices are added together.

jfgemski (approved):

tahoeblue:
Pete,

WIthout seeing the current definition, I can’t tell if you are advocating that “market rate” needs to have those terms added, or removed.  

Also, I may be misinterpreting your comments about "rate of change", and trend, but the meaning of “rate” in “market rate” is not the same meaning as “rate of change”, and analysis of “trends” is a separate function or operation that is not part of the definition of a “rate index” — which is essentially a lookup function that returns a numerical value (y-axis) for a given ‘index’ value on the x-axis (which usually is a time dimension, whether discrete or continuous). This kind of ‘rate’ is not a rate of change, or 1st derivative of a function, at that point in the indexed lookup.  It is just a returned value of the lookup function.

If I were to guess, I would think you are saying that ratio is too narrow a qualification for the numerical domain of market rate, and that the reference to "market trends” is also not appropriate.  Also, whether the value returned from a rate index lookup is a value associated with a singleton object (e.g., the price of a single share class of stock ) or is some form of composite calculation rule involving a set of objects, is irrelevant to the root definition of a “rate index” (although the distinction certainly applies when defining or describing different classes, types or individuals of rate indices.

-Jeff

Issue #1515, opened 2021-05-07
The definition of 'market rate' should not incorporate the concept of a market trend


ElisaKendall:
The definition should be revised to read 'rate that represents a ratio value established in the marketplace for' rather than 'rate that measures market trends for'

PR #1517, opened 2021-05-11
IND-107 - The definition of 'market rate' should not incorporate the concept of a market trend

ElisaKendall:

## Description

Revised the definition of market rate to read 'rate that represents a ratio value established in the marketplace for' ... per FCT discussion

Fixes: #1515 / IND-107

## Checklist:

### DIFF
--- a/IND/Indicators/Indicators.rdf
+++ b/IND/Indicators/Indicators.rdf
@@ -65,7 +65,7 @@
 		<skos:changeNote>The https://spec.edmcouncil.org/fibo/ontology/IND/20190901/Indicators/Indicators.rdf version of this ontology was modified to eliminate a redundant superclass declaration on MarketSpread, introduced by refactoring of FND analytics.</skos:changeNote>
 		<skos:changeNote>The https://spec.edmcouncil.org/fibo/ontology/IND/20200401/Indicators/Indicators.rdf version of this ontology was modified to reflect the move of dated collection from arrangements to financial dates.</skos:changeNote>
 		<skos:changeNote>The https://spec.edmcouncil.org/fibo/ontology/IND/20200701/Indicators/Indicators.rdf version of this ontology was modified to add definitions for historical and implied volatility, and differentiate price volatility accordingly.</skos:changeNote>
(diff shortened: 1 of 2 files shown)


rivettp (changes requested):
I don't think a market rate necessarily represents a ratio. SO the superclass as well as the definition should change. For example the DJIA is merely the addition of the share prices with a consistency multiplier on the total.

ElisaKendall:
@rivettp I thought you agreed to this during our meeting - I implemented what we had in the minutes.  DJIA may have a denominator of 1, but a rate is still a rate where rate is a synonym of ratio.

ElisaKendall:
@rivettp The alternative is to say that something is numeric index value, which we also have in the ontology as a direct subclass of quantity value and as a sibling of market rate but not a subclass of ratio value.  So we could change the examples for DJIA to say that it is a numeric index value rather than a market rate, if that's what you are concerned about, but a rate is a ratio by definition.

rivettp:
It's market rate that I don't think should be a ratio

ElisaKendall:
@rivettp Yes, I understand that.  The question is why -  we discussed this on our call and everyone else agreed that it should be.  A rate is a synonym of ratio.  A market rate is a rate.  One could argue that the DJIA is neither a rate nor an average, and that the value of the DJIA should be expressed as an instance of numeric index value rather than of market rate.  I would rather make that change than say that a market rate is not a rate.  Maybe I've misunderstood - are you saying that market rate is not a rate at all, but rather a synonym of 'going rate', which is a monetary amount / price? Currently our subclasses of market rate include quoted exchange rates, yield, interest rate, rate of return, etc.  These are all clearly ratio values, and their definitions correspond to the definition of market rate as we currently have it.  Some indices may have values expressed as market rates and others as numeric index values.  I think that's where we have an issue, which is that the value of the DJIA in our examples is an instance of market rate and should be an instance of numeric index value, but that doesn't mean that market rate should not be defined as a ratio (i.e., as a rate).

jfgemski (approved):

tahoeblue (approved):

dallemang (approved):

Issue #1533, opened 2021-06-03
Need the definition of a structured name and definition of a new datatype for the combination of xsd:string and rdf:langString


ElisaKendall:
Currently we represent the names of people using various strings, and similarly of organizations. But, we don't have a specific datatype allowing us to represent names as either a UTF-8 encoded name that includes language-specific characters, or as a transliterated name at the highest level. We also don't have the ability to structure names, such that the name of a person would include their family name and full legal name, for example. We also need to eliminate cultural bias from the way we represent names of people in particular, which is fraught with issues that might be construed as insensitive at best, and possibly racist at worst.

ElisaKendall:
Some of the changes we made in this issue are challenging for tools that do not support RDF 1.1.  I'm reopening this to simplify the use of the Text datatype in order to account for those issues.

PR #1534, opened 2021-06-04
FND-346  - Need the definition of a structured name and definition of a new datatype for the combination of xsd:string and rdf:langString

ElisaKendall:

## Description

This resolution introduces a new custom datatype, Text, which represents the union of xsd:string and rdf:langString, allowing better representation of names and descriptions that may be expressed in various languages.  It also adds the concept of a structured name, and augments parties, people, and organizations to use it, as a current approach to handling name patterns rather than less expressive string values, as appropriate.

Fixes: #1533 / FND-346 / FND-245 / FND-74

## Checklist:

### DIFF
--- a/BE/LegalEntities/LEIEntities.rdf
+++ b/BE/LegalEntities/LEIEntities.rdf
@@ -4,12 +4,14 @@
 	<!ENTITY fibo-be-le-fbo "https://spec.edmcouncil.org/fibo/ontology/BE/LegalEntities/FormalBusinessOrganizations/">
 	<!ENTITY fibo-be-le-lei "https://spec.edmcouncil.org/fibo/ontology/BE/LegalEntities/LEIEntities/">
 	<!ENTITY fibo-be-le-lp "https://spec.edmcouncil.org/fibo/ontology/BE/LegalEntities/LegalPersons/">
+	<!ENTITY fibo-fnd-aap-agt "https://spec.edmcouncil.org/fibo/ontology/FND/AgentsAndPeople/Agents/">
 	<!ENTITY fibo-fnd-agr-ctr "https://spec.edmcouncil.org/fibo/ontology/FND/Agreements/Contracts/">
 	<!ENTITY fibo-fnd-arr-arr "https://spec.edmcouncil.org/fibo/ontology/FND/Arrangements/Arrangements/">
 	<!ENTITY fibo-fnd-arr-cls "https://spec.edmcouncil.org/fibo/ontology/FND/Arrangements/ClassificationSchemes/">
 	<!ENTITY fibo-fnd-dt-bd "https://spec.edmcouncil.org/fibo/ontology/FND/DatesAndTimes/BusinessDates/">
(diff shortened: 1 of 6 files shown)


trypuz:
@mereolog, this PR is still WIP. Should I review it as it is?

dsnewman (approved):

dallemang (approved):
This  clearly makes a lot of sense from a modeling point of view - can you make a comment about performance, I suppose mostly in terms of Pellet evaluation of the model?  Does this make a significant change in that process?

ElisaKendall:
@dallemang - Based on the testing I've done so far, the change has had negligible impact on performance, if any, with respect to running Pellet.  Having said this, we don't have any examples of people in our set of examples in release for FIBO, so I'm not surprised about the lack of impact.  Now that this initial change is in, though, there may be some downstream updates related to organizational names that should be done, such as those that we have added in FBC for trading or operational name, as well as the need to add DBA for sole proprietorships. My intent was to put those changes off until this initial pull request was complete, and then make those changes, and see what happens.  My guess is that for the few examples that might be impacted, it will still be minimal with respect to FIBO itself.  For operational environments, there may need to be some programmatic changes to create the various name individuals associated with an organization that might add overhead in the materialization process.  Reasoning / SPARQL performance hits should be minimal though. And, I'll make sure that the existing properties can be used with or without the change to allow for their use with either a reified structural name or not.  I've already made the change to allow most of them to use the Text datatype, but the union includes xsd:string so any names in individuals that use xsd:string should still work as well.

rivettp:
This is a major architectural change and should be fully discussed rather than being pushed through over a weekend.
That's not even mentioning the significant practical and cultural problems I can foresee with the approach to naming.
Finally, putting the new datatype Text into the Agents ontology, when it is intended for widespread use, is architecturally harmful. It's already resulted in a proposed change to Markets ontology requiring a new and logically unnecessary dependency on Agents.

ElisaKendall:
This is not 'new' - you yourself reviewed this pattern in the slides that Dean and I presented at Enterprise Data World.  The revision is simply codifying that pattern for use in FIBO.  It is not mandated, however - the properties that we've specified to use the text datatype do not have a domain, and the union allows for backwards compatibility with the original use of these properties elsewhere in FIBO.

The concept of a Name as a class was added to agents because that's where it is needed for initial use, and the parties, people and organizations ontologies have already imported it and extended it in the past.  We recently simplified some of those relationships in parties, but if you look historically the relationship was there until earlier this year.  So this is nothing new.  The name class has only two elements as we discussed for a person - their full name and a surname that can be used for sorting, which is optional.  It does not include the various culturally possibly insensitive properties that remain in the ontology but are not used.  It adds an optional prefix and suffix which were a gap raised in another FND issue, however.  

There were 3 options for creation of the Text datatype: (1) put it in Agents where it is first used and needed, (2)_ put it in annotation vocabulary that everything uses, or (3) create a new tiny ontology, which you have objected to on many occasions for that purpose.  I may not have done this the way you would have, but it was a pragmatic solution based on the first place where it is used.

Issue #1532, opened 2021-06-03
The restriction on exchange that says it operates in exactly 1 municipality is too narrow


ElisaKendall:
It's possible that while an exchange is likely based in one country and within that country, in one municipality, it is possible given today's digital trading environment that it could operate in more than one location (even though the current set of MIC codes force a choice of a single location).

One example is NYSE Arca, whose parent organization is in New York, but that operates, or has operated in Chicago. Because our example conflicts with the MIC code representation, Chicago is inferred to be the same as New York, which is clearly not right.

jfgemski:
I think the distinction you want to make is “primarily resides in” (the MIC code) and “operates in”.

 

John

 

***@***.*** ***@***.***>  |O: (516) 826-0701| M: (516) 353-0004

 

From: Elisa Kendall ***@***.***> 
Sent: Thursday, June 3, 2021 2:56 PM
To: edmcouncil/fibo ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [edmcouncil/fibo] The restriction on exchange that says it operates in exactly 1 municipality is too narrow (#1532)

 

It's possible that while an exchange is likely based in one country and within that country, in one municipality, it is possible given today's digital trading environment that it could operate in more than one location (even though the current set of MIC codes force a choice of a single location).

One example is NYSE Arca, whose parent organization is in New York, but that operates, or has operated in Chicago. Because our example conflicts with the MIC code representation, Chicago is inferred to be the same as New York, which is clearly not right.

PR #1536, opened 2021-06-07
FBC-281 - The restriction on exchange that says it operates in exactly 1 municipality is too narrow

ElisaKendall:

## Description

Modified the markets ontology to loosen constraints on the location in which a given exchange operates, given that there are cases when an exchange may operate in multiple locations, and to allow an exchange name to be multilingual or expressed in a language other than English

Fixes: #1532 / FBC-281

## Checklist:


rivettp (changes requested):
I don't like the idea of the extra dependency on Agents. Why is something so generic as Text defined there?

ElisaKendall:
@rivettp Should it be in its own ontology with only the datatype and related property definition?  Otherwise agents seemed the logical place to include it as the first place it is used.

ElisaKendall:
@rivettp I've removed the dependency on agents per your request, and will further simplify the change related to the Text datatype in a new update to the issue raised in #1533, so that we avoid some of the challenges in Protege / Pellet, but allow people to use the new datatype as needed in applications that understand RDF 1.1.  I'm currently testing that in a couple of other tools.

jfgemski (approved):

mereolog:
@rivettp given [Elisa's change](https://github.com/edmcouncil/fibo/pull/1536#issuecomment-857870843) would you mind re-reviewing this PR?

PR #1671, opened 2022-01-04
SEC-170 - Further work is needed to clean up the pool-backed securities ontology

ElisaKendall:

## Description

This resolution adds the concept of an instrument pool as an asset to link it to its ownership; moves the concept of an SPV from Funds to Pools so that it is more broadly accessible; adds the concept of an issuer to the definition of structured finance instrument and addresses other issues raised, such as revision of definitions and additional restrictions on several concepts.

Fixes: #1671 / SEC-170

## Checklist:


jfgemski (dismissed):

jfgemski (approved):

jfgemski (approved):

SEC/Debt/AssetBackedSecurities/MortgageBackedSecurities.rdf
```diff
@@ -127,7 +127,7 @@
 	</owl:Class>
 	
 	<owl:Class rdf:about="&fibo-sec-dbt-ab-mbs;AgencyMBSDeal">
-		<rdfs:subClassOf rdf:resource="&fibo-sec-dbt-ab-mbs;MortgageBackedSecurityOffering"/>
+		<rdfs:subClassOf rdf:resource="&fibo-sec-dbt-dbti;DebtOffering"/>
 		<rdfs:label xml:lang="en">agency m b s deal</rdfs:label>
```
rivettp, line 131: We should use MBS in the label

SEC/Debt/AssetBackedSecurities/MortgageBackedSecurities.rdf
```diff
@@ -449,7 +437,7 @@ Consensus:Review.</skos:editorialNote>
 	</owl:Class>
 	
 	<owl:Class rdf:about="&fibo-sec-dbt-ab-mbs;PassThroughMBSDeal">
```
rivettp, line 439: We need to revisit the whole term "deal" which I cannot see used outside FIBO. "Deal" seems to imply "contract" which si not at all the same as an offering which does not become a contract until the offer is taken/purchased

SEC/Debt/AssetBackedSecurities/PoolBackedSecurities.rdf
```diff
@@ -35,11 +37,11 @@
...
-		Note that in common industry parlance, the term &apos;asset backed security&apos; is used to refer to securities which specifically do not have a mortgage loan pool as their underlying asset. For this reason, this ontology provides a common parent to both asset backed and mortgage backed securities, which is effectively an asset backed security in the broader sense. This has been labeled as &apos;pool backed security instrument&apos; as there is no common industry name for this common parent class, other than &apos;asset backed security&apos;.</dct:abstract>
+		<dct:abstract>This ontology defines concepts that are common to asset-backed and mortgage-backed securities, including pools, as well as structured finance instruments. 
+		Note that in common industry parlance, the term &apos;asset backed security&apos; is used to refer to securities which specifically do not have a mortgage loan pool as their underlying asset. For this reason, this ontology provides a common parent to both asset-backed and mortgage-backed securities, i.e., a pool-backed security, which is effectively an asset backed security in the broader sense.</dct:abstract>
```
rivettp, line 41: this seems to be saying that our semantic model is based not on semantics but a judgement of common industry parlance, which is a confusing message. Have we applied that consistently? 
Seems we could avoid confusion by avoiding the confusing names as primary class/property names and keep them as synonyms, with more neutral names in the ontology. In this case "Pool backed" is good but we should consider renaming "asset backed" to "NonMortgageLoanBackedSecurity": a bit unwieldy I know but it reflects the semantics.
Either way we should document our approach more generally in contribution guidelines.

SEC/Debt/AssetBackedSecurities/PoolBackedSecurities.rdf
```diff
@@ -109,21 +93,25 @@
 	<owl:Class rdf:about="&fibo-sec-dbt-ab-pbs;PrincipalProtectedNote">
 		<rdfs:subClassOf rdf:resource="&fibo-sec-dbt-ab-pbs;StructuredFinanceInstrument"/>
 		<rdfs:label xml:lang="en">principal protected note</rdfs:label>
-		<skos:definition xml:lang="en">structured finance that offers investors exposure to chosen underlying assets using various approaches and offering a large variety of asymmetric pay-off profiles</skos:definition>
+		<skos:definition xml:lang="en">structured finance that offers investors exposure to chosen underlying assets using various approaches and asymmetric pay-off profiles</skos:definition>
```
rivettp, line 96: this doesn't differentiate from any other structured finance instrument nor bring in the aspect of principal protection. The class name seems too general - I'm sure there are other principal protected notes that are not structured finance e.g., inflation rate bonds

SEC/Debt/AssetBackedSecurities/PoolBackedSecurities.rdf
```diff
@@ -109,21 +93,25 @@
...
+					<owl:Restriction>
+						<owl:onProperty rdf:resource="&fibo-fnd-pty-rl;isPlayedBy"/>
+						<owl:someValuesFrom rdf:resource="&fibo-sec-sec-pls;SpecialPurposeVehicle"/>
+					</owl:Restriction>
+				</owl:someValuesFrom>
+			</owl:Restriction>
+		</rdfs:subClassOf>
 		<rdfs:label xml:lang="en">structured finance instrument</rdfs:label>
 		<rdfs:seeAlso rdf:resource="https://www.bis.org/publ/cgfs23mitchell.pdf"/>
 		<skos:definition xml:lang="en">pool-backed security wherein the risk associated with the pool has been uncoupled from the risk associated with the originating institution, typically through a special purpose vehicle</skos:definition>
```
rivettp, line 117: "typically" is wrong with respect to the new someValuesFrom restriction

SEC/Debt/AssetBackedSecurities/PoolBackedSecurities.rdf
```diff
@@ -135,6 +123,7 @@
 	<owl:Class rdf:about="&fibo-sec-dbt-ab-pbs;StructuredFinanceWithoutCapitalProtection">
 		<rdfs:subClassOf rdf:resource="&fibo-sec-dbt-ab-pbs;StructuredFinanceInstrument"/>
 		<rdfs:label xml:lang="en">structured finance without capital protection</rdfs:label>
+		<owl:disjointWith rdf:resource="&fibo-sec-dbt-ab-pbs;PrincipalProtectedNote"/>
 		<skos:definition xml:lang="en">structured finance that is a short-term note linked to an underlying asset that offers a steady stream of income</skos:definition>
```
rivettp, line 127: definition seems adrift from the class name which is about no protection. 
Also "Without Capital Protection" and "Principal Protected" are inconsistent.
 Why does it have to be "short term"?

SEC/Securities/Pools.rdf
```diff
@@ -200,5 +230,13 @@
 		<skos:definition>security that is included in a securities pool</skos:definition>
 		<fibo-fnd-utl-av:explanatoryNote>This is a pool of any kind of security, and not therefore necessarily a pool of debt, though usually it is debt securities which are pooled.</fibo-fnd-utl-av:explanatoryNote>
 	</owl:Class>
+	
+	<owl:Class rdf:about="&fibo-sec-sec-pls;SpecialPurposeVehicle">
```
rivettp, line 234: I'm wondering if this is still not general enough. Maybe it should be in BE. Can't SPVs be created for other reasons than related to Pools? For example from Investopedia
What Is the Function of SPVs in Public-Private Partnerships?
Public-private partnerships are collaborations between a government agency and a privately owned company. Many private partners in public-private partnerships demand a special purpose vehicle as part of the arrangement. This is especially true for capital-intensive endeavors, such as an infrastructure project. The private company might not want to take on too much financial exposure, so an SPV is created to absorb some of the risks

rivettp (changes requested):
There are a few issues. In particular I think we should move SPV to BE before merging thsi fairly disruptive change.

SEC/Debt/AssetBackedSecurities/MortgageBackedSecurities.rdf
ElisaKendall: @rivettp - I agree, but didn't want to make more changes via this issue than necessary.

SEC/Debt/AssetBackedSecurities/PoolBackedSecurities.rdf
ElisaKendall: @rivettp We could revise the abstract, but I'm against revising the name at this stage.  I'll remove the comment from the abstract since I don't think it is necessary.  It was 'left over' from the original as you can see.

SEC/Debt/AssetBackedSecurities/PoolBackedSecurities.rdf
ElisaKendall: @rivettp Ok - I'll see if I can find differentiating characteristics to include in the definition, but that may require assistance from JG.

SEC/Debt/AssetBackedSecurities/PoolBackedSecurities.rdf
ElisaKendall: @rivettp - removed 'typically'

SEC/Debt/AssetBackedSecurities/PoolBackedSecurities.rdf
ElisaKendall: @rivettp - that is the definition from ISO, which we can revisit if we have a better source, but I changed the name to be consistent with principal-protected and added the synonym of capital protected

SEC/Securities/Pools.rdf
ElisaKendall: @rivettp - ok, I'll move it again.  That crossed my mind when I put it in pools, which is significantly higher up the food chain than where it started.

SEC/Debt/AssetBackedSecurities/PoolBackedSecurities.rdf
rivettp: If we do leave in "short term" we should reflect in some sort of restriction. And consider which other instruments are invariably "short term" including bonds. Maybe define the mixin class "short term debt instrument"? As well as other terms.

SEC/Debt/AssetBackedSecurities/PoolBackedSecurities.rdf
rivettp: I'm less worried about the ontology name (to which the abstract applies) than the class name which is what my suggestion was intended to be for.

SEC/Debt/AssetBackedSecurities/MortgageBackedSecurities.rdf
rivettp: Agreed - it is a separate issue and I'm OK with deferring.

jfgemski (approved):

SEC/Debt/AssetBackedSecurities/MortgageBackedSecurities.rdf
ElisaKendall: @rivettp I think this may go away altogether and there are many cases in the MBS ontology of the misuse of abbreviations that we have yet to address.

SEC/Debt/AssetBackedSecurities/PoolBackedSecurities.rdf
ElisaKendall: @rivettp I think the current name is what is used in the industry, and that there is some overloading of the concept we have called pool-backed security which some also refer to as asset-backed security, but the narrower concept is correctly named.  We can confirm that with John and Jeff in two weeks, or on another call, but I would think that we would have egg on our face if we did not have a definition for asset-backed security.

rivettp (approved):

jfgemski (approved):

Issue #1673, opened 2022-01-04
Merge the IROptions ontology in rate derivatives with ExoticOptions in the main derivatives contracts module


ElisaKendall:
The contents of the IROptions ontology should be merged with the ExoticOptions ontology, which includes caps and floors, and the IROptions ontology itself should be eliminated.

PR #1677, opened 2022-01-10
DER-120 - Merge the IROptions ontology in rate derivatives with ExoticOptions in the main derivatives contracts module

ElisaKendall:

## Description

Merged IROptions with ExoticOptions and augmented the Debt ontology with the concept of interest rat cap, which was included in the IROptions ontology and was a gap

Fixes: #1673 / DER-120

## Checklist:


jfgemski (dismissed):

DER/DerivativesContracts/ExoticOptions.rdf
```diff
@@ -150,6 +150,23 @@
...
+		<rdfs:label xml:lang="en">cap loan</rdfs:label>
+	</owl:Class>
+	
+	<owl:Class rdf:about="&fibo-der-der-exo;CapOrFloorDetermination">
+		<rdfs:subClassOf rdf:resource="&fibo-fnd-gao-obj;Strategy"/>
+		<rdfs:label xml:lang="en">cap or floor determination</rdfs:label>
+		<skos:definition xml:lang="en">Whether a given thing is treated as a Cap (something goes no higher than this amount) or as a Floor (something goes no lower than this).</skos:definition>
+	</owl:Class>
+	
 	<owl:Class rdf:about="&fibo-der-der-exo;CashOrNothingPayoutCommitment">
```
rivettp, line 170: This class is not needed: we don't model other required payments as "commitments"

FBC/DebtAndEquities/Debt.rdf
```diff
@@ -628,6 +636,12 @@
 		<fibo-fnd-utl-av:explanatoryNote>Note that in most cases, the dates and payment frequencies for interest will coincide with the dates and payment frequencies related to the principal.</fibo-fnd-utl-av:explanatoryNote>
 	</owl:Class>
 	
+	<owl:Class rdf:about="&fibo-fbc-dae-dbt;InterestRateCap">
```
rivettp, line 639: This does not seem right as a class: surely the only class needed is InterestRate and whether it's treated as a cap or floor should be determined by how it's used in properties e.g. hasInterestRateCap.

FBC/DebtAndEquities/Debt.rdf
```diff
@@ -951,6 +965,13 @@
 		<skos:definition>relates something, such as an agreement, or debt instrument, to the rate (typically annual) of interest that is to be paid by the debtor to the creditor on the debt</skos:definition>
 	</owl:ObjectProperty>
 	
+	<owl:ObjectProperty rdf:about="&fibo-fbc-dae-dbt;hasInterestRateCap">
+		<rdfs:subPropertyOf rdf:resource="&fibo-fbc-dae-dbt;hasInterestRate"/>
+		<rdfs:label>has interest rate cap</rdfs:label>
+		<rdfs:range rdf:resource="&fibo-fbc-dae-dbt;InterestRateCap"/>
```
rivettp, line 971: It's redundant to have a specific class here - just use InterestRate.

rivettp (changes requested):
I guess the removal of IR Swaps as an ontology makes sense, but the Exotic Options ontology remains a mess.
The reason I've requesting a change is I'm not happy with InterestRateCap as a specific subclass of InterestRate, and this is in a Release ontology. It's like having a separate subclass of Percentage or CurrencyAmount each time they're used in a different property.

DER/DerivativesContracts/ExoticOptions.rdf
```diff
@@ -150,6 +150,23 @@
...
+	</owl:Class>
+	
+	<owl:Class rdf:about="&fibo-der-der-exo;CapLoan">
+		<rdfs:subClassOf rdf:resource="&fibo-fbc-fi-fi;DebtInstrument"/>
+		<rdfs:label xml:lang="en">cap loan</rdfs:label>
+	</owl:Class>
+	
+	<owl:Class rdf:about="&fibo-der-der-exo;CapOrFloorDetermination">
+		<rdfs:subClassOf rdf:resource="&fibo-fnd-gao-obj;Strategy"/>
```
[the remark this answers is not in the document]
ElisaKendall: @rivettp The approach I took was to move the content from the IROptions ontology directly across, without any change.  We have another open issue regarding cleaning up Exotic Options, which I believe will take several weeks if not longer, and I suggest that we address this (if we need it at all), at that time.

DER/DerivativesContracts/ExoticOptions.rdf
ElisaKendall: @rivettp See my comment above about not making changes to the content of the IROptions ontology, simply moving it straight across, for this particular pull request.

FBC/DebtAndEquities/Debt.rdf
ElisaKendall: @rivettp The reason that I created a class for this is that financial institutions are required to report on this specific item as a component of their regulatory reports monthly to a myriad of regulators.  It also factors into all kinds of analyses related to cash flows / earnings over time.  Otherwise I would have modeled it as you suggest.  Based on my experience at WF, it is far easier to map concepts like this to various repos if there is a class to map to rather than limiting the model to use of a property.

FBC/DebtAndEquities/Debt.rdf
ElisaKendall: @rivettp See my comment above regarding the fact that this is a required field for regulatory reporting, and that it can be difficult to map across repos without a class to map to - that allows for format inconsistencies across repos for example, which use of a single property makes more difficult.

jfgemski (approved):

tahoeblue (approved):

Issue #1766, opened 2022-04-19
Need the ability to model composite identifiers in FIBO


ElisaKendall:
Integrate the ability to incorporate regular expressions with the concept of a composite identifier to allow for support of the UTI properly. The UTI is composed of an LEI plus an internal scheme that uniquely identifies the transaction. The scheme for minting the identifier includes an LEI plus some other opaque transaction identifier according to the ISO standard.

PR #1767, opened 2022-04-19
DER-129 - Need the ability to model composite identifiers in FIBO

ElisaKendall:

## Description

Revised the definition of a composite identifier to (1) loosen the restriction on having at least one identifier to optionally having an identifier (as it could be a code), and (2) added two annotations for constructing and deconstructing a regular expression that states precisely how to build the composite identifier

Fixes: #1766 

## Checklist:

### DIFF
--- a/FBC/FunctionalEntities/FinancialServicesEntities.rdf
+++ b/FBC/FunctionalEntities/FinancialServicesEntities.rdf
@@ -72,8 +72,8 @@
 		<dct:abstract>This ontology defines basic financial service providers, such as holding companies, financial institutions (both depository and non-depository institutions), and clearing houses at a relatively general level. Nuances specific to the institutions located in a specific country are defined in jurisdiction specific dependent ontologies.</dct:abstract>
 		<dct:license rdf:datatype="&xsd;anyURI">http://opensource.org/licenses/MIT</dct:license>
 		<sm:contentLanguage rdf:resource="https://www.w3.org/TR/owl2-quick-reference/"/>
-		<sm:copyright>Copyright (c) 2015-2021 EDM Council, Inc.</sm:copyright>
-		<sm:copyright>Copyright (c) 2015-2021 Object Management Group, Inc.</sm:copyright>
+		<sm:copyright>Copyright (c) 2015-2022 EDM Council, Inc.</sm:copyright>
+		<sm:copyright>Copyright (c) 2015-2022 Object Management Group, Inc.</sm:copyright>
 		<sm:dependsOn rdf:datatype="&xsd;anyURI">https://spec.edmcouncil.org/fibo/ontology/BE/</sm:dependsOn>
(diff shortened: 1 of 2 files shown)


jfgemski (approved):

tahoeblue:
I think a clearer definition of "identifier" needs to be made that distinguishes between an identifier that is opaque (i.e., without any 'intelligence' or constructed components), and what I would call "composite identification schemes" in which the no-intelligence restriction is relaxed, and facets from multiple sources are used to "construct" an identifier.  
The "fictional" example of a RegistrationAuthorithy (RA) concatenated with a regex-based string rule may fit a subset of how legacy identifiers have been used, but to the extent that it is a suggestion as to a general definition or means by which a composite identification scheme can be implemented, I think it is too narrow.
It is more appropriate to document how some frequently used financial identification schemes have been deployed (such as BIC and ISIN) without trying to have a general class for such schemes, as any such scheme is essentially unique in how it has been designed.
Specifying a rule for what constitutes the valid *format* of an identifier (allowable characters, positions, checksums, etc) in order to validate that an identifier (such as a credit card number) is valid is one thing.   But to infer semantic meaning on the so-called components of identifiers as part of a definition of a "composite identifier" sets a bad precedent, in my opinion.
For example, the partitioning of the LEI address space with prefixes that are allocated to LEI registrars has been used as an example of a "composite index" following the fule of "RA:regex", where RA = a prefix assigned to an LOU (LEI registrar).  However, the assumption that a prefix identifies an LEI registrar is not correct, for two reasons:
(1) If an LEI registration is transferred from one LOU to another, the prefix is not updated with a prefix from the recipient LOU (an LEI is not only unique but persistent -- once minted, it never changes); and
(2) There are LEIs that exist with no such allocation of prefixes (from a period in which the entire LEI was generated as a random number.
Again, I would prefer that this concept be referred to as a "composite identifier scheme", and not a "composite identifier" (a concept that I think introduces misconceptions regarding the interpretation of "identifier" )

ElisaKendall:
@tahoeblue One of the things that we added were two annotation properties to allow specification of the various regex expressions that might be appropriate - we agree with you on that.

There are other cases outside of the finance community where the notion of a composite identifier is clearer, such as in Pharmaceuticals, where the identifier includes a country code and language code in some cases.  If a particular substance is identified for use by the FDA, it may have an identifier including 'US', and if in Europe and approved by EMA it may include 'EU'.  Those are really the kinds of things we are talking about here for the most part.  The idea is to pre-pend such codes so that the identifiers are globally unique, since they may be managed independently, unlike the LEI.

tahoeblue:
All the more reason to call such a practice a "scheme" and not an "identifier".  An identifier (in a particular name space) is administered and managed in order to be unique per the identification rules of the space.  Without a governing body to administer the type of country-code-prepended scheme that you mention above, the scheme may not itself be able to guarantee uniqueness, or be more than a composite database key convention (which is fine, of course) -- just not an actual, standardized identification regime.

Issue #1791, opened 2022-06-29
Need to better integrate the concept of a UTI conforming to the ISO standard


ElisaKendall:
There is a definition of a UTI in FIBO, as a subclass of TradeIdentifier. We also have transaction identifier and securities transaction identifier, but none of them are well integrated with one another. This needs reconciliation.

PR #1792, opened 2022-06-30
DER-125 - Need to better integrate the concept of a UTI conforming to the ISO standard

ElisaKendall:

## Description

Reconciled (deprecated) securities transaction identifier in favor of unique transaction identifer, adjusted semantics and definitions to better align with the ISO standard

Fixes: #1791 

## Checklist:

### DIFF
--- a/FBC/ProductsAndServices/FinancialProductsAndServices.rdf
+++ b/FBC/ProductsAndServices/FinancialProductsAndServices.rdf
@@ -82,8 +82,8 @@
 		<dct:abstract>This ontology defines concepts that extend the Foundations (FND) Products and Services concepts specifically for the financial industry, including financial product, financial service, and financial service provider.</dct:abstract>
 		<dct:license rdf:datatype="&xsd;anyURI">http://opensource.org/licenses/MIT</dct:license>
 		<sm:contentLanguage rdf:datatype="&xsd;anyURI">https://www.w3.org/TR/owl2-quick-reference/</sm:contentLanguage>
-		<sm:copyright>Copyright (c) 2015-2021 EDM Council, Inc.</sm:copyright>
-		<sm:copyright>Copyright (c) 2015-2021 Object Management Group, Inc.</sm:copyright>
+		<sm:copyright>Copyright (c) 2015-2022 EDM Council, Inc.</sm:copyright>
+		<sm:copyright>Copyright (c) 2015-2022 Object Management Group, Inc.</sm:copyright>
 		<sm:dependsOn rdf:datatype="&xsd;anyURI">https://spec.edmcouncil.org/fibo/ontology/BE/</sm:dependsOn>
 		<sm:dependsOn rdf:datatype="&xsd;anyURI">https://spec.edmcouncil.org/fibo/ontology/FBC/FunctionalEntities/RegulatoryAgencies/</sm:dependsOn>
 		<sm:dependsOn rdf:datatype="&xsd;anyURI">https://spec.edmcouncil.org/fibo/ontology/FND/</sm:dependsOn>
@@ -132,6 +132,7 @@
 		<skos:changeNote>The https://spec.edmcouncil.org/fibo/ontology/FBC/20210301/ProductsAndServices/FinancialProductsAndServices/ version of this ontology was modified to rename ownership related properties for consistent alignment with the ownership situational pattern, add a definition for trading strategy, and loosen the constraint on offeree for offering to be optional.</skos:changeNote>
 		<skos:changeNote>The https://spec.edmcouncil.org/fibo/ontology/FBC/20210601/ProductsAndServices/FinancialProductsAndServices/ version of this ontology was modified to add a property to describe the criteria for including something in a basket, if that criteria is known, and to point to a party that is responsible for determining that criteria.</skos:changeNote>
 		<skos:changeNote>The https://spec.edmcouncil.org/fibo/ontology/FBC/20210901/ProductsAndServices/FinancialProductsAndServices/ version of this ontology was modified to fix spelling errors.</skos:changeNote>
+		<skos:changeNote>The https://spec.edmcouncil.org/fibo/ontology/FBC/20211201/ProductsAndServices/FinancialProductsAndServices/ version of this ontology was modified to revise the definition of a unique transaction identifier to align with ISO 23897.</skos:changeNote>
 		<fibo-fnd-utl-av:hasMaturityLevel rdf:resource="&fibo-fnd-utl-av;Release"/>
 	</owl:Ontology>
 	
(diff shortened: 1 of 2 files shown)


jfgemski (approved):

FBC/ProductsAndServices/FinancialProductsAndServices.rdf
```diff
@@ -785,12 +786,12 @@
...
 				<owl:onProperty rdf:resource="&lcc-lr;identifies"/>
-				<owl:onClass rdf:resource="&fibo-fbc-pas-fpas;Trade"/>
-				<owl:qualifiedCardinality rdf:datatype="&xsd;nonNegativeInteger">1</owl:qualifiedCardinality>
+				<owl:someValuesFrom rdf:resource="&fibo-fbc-pas-fpas;Trade"/>
 			</owl:Restriction>
 		</rdfs:subClassOf>
 		<rdfs:label>trade identifier</rdfs:label>
-		<skos:definition>an identifier for a trade</skos:definition>
+		<skos:definition>sequence of characters identifying a trade within some context</skos:definition>
+		<skos:note>Note that a given trade may consist of multiple transactions, and thus there may be multiple UTIs associated with a trade.</skos:note>
```
rivettp, line 794: have not mentioned UTI in the context of this class

FBC/ProductsAndServices/FinancialProductsAndServices.rdf
```diff
@@ -985,15 +986,16 @@
 		<rdfs:subClassOf rdf:resource="&fibo-fbc-pas-fpas;TradeIdentifier"/>
```
rivettp, line 986: we say earlier that a Trade may consist of many Transactions (UTIs) and so saying UTI is a subclass of TradeIdentifier does not seem consistent

FBC/ProductsAndServices/FinancialProductsAndServices.rdf
```diff
@@ -985,15 +986,16 @@
...
 		<rdfs:subClassOf>
 			<owl:Restriction>
-				<owl:onProperty rdf:resource="&fibo-fbc-pas-fpas;hasGeneratingEntityIdentifier"/>
-				<owl:someValuesFrom rdf:resource="&fibo-be-le-lei;LegalEntityIdentifier"/>
+				<owl:onProperty rdf:resource="&fibo-fbc-pas-fpas;hasGeneratingEntity"/>
```
rivettp, line 989: we don't a good job of defining "generated". And what's the significance of the entity. If market maker A invokes a service run by provider B using code written by software company C to generate a UTI then which entity is said to have "generated" it?

rivettp (changes requested):
a few comments. I'm most worried about UTI being a subclass of TradeIdentifier which is inconsistent with us saying a Trade may consist of many Transactions/UTIs

tahoeblue (approved):

FBC/ProductsAndServices/FinancialProductsAndServices.rdf
ElisaKendall: It is.  In many cases there is only one transaction, in which case they would be the same identifier.  It's only in cases where there are multiple transactions comprising a trade when they are many to 1.

FBC/ProductsAndServices/FinancialProductsAndServices.rdf
jfgemski: ISO must have an old definition that doesn't account for program trading.  In program trading of course the UTI is programmatically generated.  For a manually entered trade, the trader uses a pre-printed trade ticket that has the UTI pre-assigned.  Thus it is right that the entity assigns the UTI in both cases.  In both cases, each trade is one transaction.  In the case of program trading, multiple trades are created in what's known as a trade list.  But here again, each trade in the trade list corresponds to one transaction.

ElisaKendall: @jfgemski The ISO specification is from 2020, so really recent.  I can add this note, but are you certain that in every case there is a 1-1 correspondence between the definition of what a trade is and what a transaction is?  That wasn't the case at Nordea - they would execute multiple transactions as part of a 'trade', which might include more than one counterparty.  They also had an overarching notion of a 'deal', and there might be additional internal hedge transactions that would be considered part of the deal under an overall identifier for a given trade, and then the various transactions would be linked to that overarching deal/trade.

ElisaKendall: @jfgemski Our definition of trade is: "an agreement between parties participating in a voluntary action of buying and selling goods and services" - very general on purpose, though it came from several sources.  There are a number of notes associated with it as well.

In a subsequent issue, we should probably add the notion of  'program trade', and possibly take another look at the lifecycle definitions in that ontology for the lifecycle of a trade.  I have at least one or two references that may be useful.  But the intent of this issue was to eliminate some schizophrenia with multiple securities transaction identifier / unique transaction identifier concepts that were disconnected.  I think this resolves that issue, even if we determine there is still some ambiguity to resolve.

jfgemski: What we'll find is that different entities have different definitions of what they consider to be a trade and a deal.  Nordea's concept of a trade can have multiple transactions.  (I wonder what they think is the difference between a trade and a deal.)  In any case, each transaction has its own UTI.  Let's close this.

rivettp:
The aspect of generating entity is covered in depth in section 3.3 of the following https://www.bis.org/cpmi/publ/d158.pdf (though this is not the definitive work) and it's  about the entity identifier embedded into the UTI itself - in other words entity B in my example. Hence having a separate property for the generating entity seems somewhat redundant, though I'm not against it. We should however add a note to explain what it's for.

tahoeblue:
Pete,

I agree with the jursidictional dependency on when “a transaction is required to be reported”, but I still think that the “transaction” mentioned in this language is the same thing as “trade”, not the intermediate processing steps performed by market infrastructures to complete the transaction/trade.   If nothing else, the need for every operational processor to generate a UTI for a subsequent interaction with another market infrastructure processor in the course of completing an exchanged-based trade would swamp the processing of trades with counterproductive complexity and volume of UTI creation without any need or requirement to report them.

FBC/ProductsAndServices/FinancialProductsAndServices.rdf
ElisaKendall: @tahoeblue Thanks Jeff, That's how I understood it.  The places in Nordea's process where there might be more than one UTI involved in the 'trade' side of a deal (the customer facing side) seemed to involve either multiple bonds (one of which might be zero coupon) or structured products, or other more complex deals, some of which involved more than one counterparty.  They also had an internal "bank side" of a deal, where they hedged risk internally, but it was all part of the same overarching "deal".  There were many more internal steps that would not count as separate "transactions" as you've outlined.

FBC/ProductsAndServices/FinancialProductsAndServices.rdf
ElisaKendall: @rivettp This last statement with respect to individual components of a package (which might reflect a structured product or something like that, at least it did at Nordea), is why internally they needed to track (1) an internal trade identifier, and (2) one or in this case more than one associated transaction-specific identifier.  But that's not inconsistent with a UTI being a subclass of trade identifier due to the restrictions we have now.

ElisaKendall:
> The aspect of generating entity is covered in depth in section 3.3 of the following https://www.bis.org/cpmi/publ/d158.pdf (though this is not the definitive work) and it's about the entity identifier embedded into the UTI itself - in other words entity B in my example. Hence having a separate property for the generating entity seems somewhat redundant, though I'm not against it. We should however add a note to explain what it's for.

@rivettp In addition to the restriction on the UTI relating to the legal entity, we also have the property chain that links to the LEI of that legal entity which can be used for this purpose.  The restriction relating to the legal entity as the generating entity is in the latest version of the ISO standard, though, which I purchased so we would have it.

Issue #1950, opened 2023-07-16
fibo-be-ge-ukj:EnglandAndWalesJurisdiction references missing resource


aamedina:
fibo-be-ge-ukj:EnglandAndWalesJurisdiction declares fibo-fnd-law-jur:hasReach
lcc-3166-2-gb:EnglandAndWales, which I could not find. So in this branch I replaced it with two references to both England and Wales individually.

https://github.com/edmcouncil/fibo/compare/master...aamedina:edm-fibo:ukj

PR #1953, opened 2023-07-23
issue-1950: add separate references to england and wales

aamedina:
## Description

fibo-be-ge-ukj:EnglandAndWalesJurisdiction declares fibo-fnd-law-jur:hasReach
lcc-3166-2-gb:EnglandAndWales, which I could not find. So in this branch I replaced it with two references to both England and Wales individually.

Fixes: #1950 

## Checklist:

### DIFF
--- a/BE/GovernmentEntities/EuropeanJurisdiction/UKGovernmentEntitiesAndJurisdictions.rdf
+++ b/BE/GovernmentEntities/EuropeanJurisdiction/UKGovernmentEntitiesAndJurisdictions.rdf
@@ -53,10 +53,11 @@
 		<owl:imports rdf:resource="https://www.omg.org/spec/LCC/Countries/ISO3166-1-CountryCodes/"/>
 		<owl:imports rdf:resource="https://www.omg.org/spec/LCC/Countries/ISO3166-2-SubdivisionCodes/"/>
 		<owl:imports rdf:resource="https://www.omg.org/spec/LCC/Countries/Regions/ISO3166-2-SubdivisionCodes-GB/"/>
-		<owl:versionIRI rdf:resource="https://spec.edmcouncil.org/fibo/ontology/BE/20230301/GovernmentEntities/EuropeanJurisdiction/UKGovernmentEntitiesAndJurisdictions/"/>
+		<owl:versionIRI rdf:resource="https://spec.edmcouncil.org/fibo/ontology/BE/20230701/GovernmentEntities/EuropeanJurisdiction/UKGovernmentEntitiesAndJurisdictions/"/>
 		<skos:changeNote>The http://www.omg.org/spec/EDMC-FIBO/BE/20200201/GovernmentEntities/EuropeanJurisdiction/UKGovernmentEntitiesAndJurisdictions.rdf version of this ontology was modified to add devolved government entities for Scotland, Wales, and Northern Ireland.</skos:changeNote>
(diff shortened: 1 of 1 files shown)


rivettp (changes requested):
Good catch but the fix is not correct: the URIs for England and Wales are as follows, based on LCC 1.2:
&lcc-3166-2-gb;GB-ENG-Subdivision
&lcc-3166-2-gb;GB-WLS-Subdivision

rivettp:
@mereolog this PR passed hygiene tests despite using a non-existent reference to LCC &lcc-3166-2-gb;England. Can those not be checked? Either by live access or by using a cached copy of LCC (it rarely changes).

aamedina:
The source of the model I used as reference is here: https://www.omg.org/spec/LCC/Countries/Regions/ISO3166-2-SubdivisionCodes-GB.rdf

which contains the England & Wales aliases (see: https://github.com/aamedina/fibo/blob/07446cba68193d52d15598cbf59a76bbbca3508a/src/cljc/net/wikipunk/rdf/lcc_3166_2_gb.cljc#L421) with a owl:sameAs relation to &lcc-3166-2-gb;GB-ENG-Subdivision. I will update the patch to reflect these identities tomorrow.

thank you for the review!

ElisaKendall:
@aamedina Hi Adrian, I still use the aliases - Pete implemented the code based subdivisions for a number of reasons, including but not limited to ease of use for programmers. However, the aliases will remain due to requirements for them by business users in a number of domain areas. I'm actually fine with the way that you had it, but our rules state that we need two approvals.

Note that there are metadata changes I would like to see: (1) add a change note, referencing this issue and stating what was changed, and (2) revise the version IRI to '20230701'.

Thanks! 

Elisa

mereolog:
> @mereolog this PR passed hygiene tests despite using a non-existent reference to LCC &lcc-3166-2-gb;England. Can those not be checked? Either by live access or by using a cached copy of LCC (it rarely changes).

We do not check resources outside the 'edmcouncil' domain.
As far as I can tell we never did - probably because we do not have governance over "external resources".

I think I can add an additional check where all resources are checked for their types - but it will produce just warnings. Otherwise, I suspect that we shortly will have to turn it off because it will block the ontology development.

ElisaKendall (approved):

jfgemski (approved):

aamedina:
I think this just another review from @rivettp? Anything on my end that needs to happen?

ElisaKendall:
> I think this just another review from @rivettp? Anything on my end that needs to happen?

Hi Adrian, if you would please merge the master branch back into this one, it should (in theory) run back through all of the checks. If Pete doesn't have time to re-approve this, @mereolog can do so. I think this is the last issue we will be addressing before the 2023 Q3 release, FYI. Thanks for checking!

aamedina:
Hmm, can you try re-running the build? Looks like there was an issue with pulling the docker image in Jenkins. Maybe cold booting? 

```shell
docker pull "$JD_TO_PULL"— Shell Script10s

+ docker pull edmcouncil/ontology-publisher:stable

Error response from daemon: Head "https://registry-1.docker.io/v2/edmcouncil/ontology-publisher/manifests/stable": received unexpected HTTP status: 503 Service Unavailable

script returned exit code 1
```

rivettp (approved):
It would be a separate issue but it's a bit odd that we have JurisdictionOfEnglandAndWales governed only by the Welsh Assembly. In fact that Assembly only has limited powers with the UK Parliament having the majority of control (the Scottish Assembly has more, as indicated by having a separate jurisdiction.). I don't think we're a Governmental KG so don't think we should get into this nuance. The key question is what are the use cases for separating jurisdiction from geographical region in the first place?

Issue #2021, opened 2024-06-24
Deprecated resources in use


mereolog:
There are three deprecated resources still in use:
1. https://spec.edmcouncil.org/fibo/ontology/FND/Relations/Relations/hasIdentity
2. https://spec.edmcouncil.org/fibo/ontology/SEC/Securities/SecuritiesIdentificationIndividuals/ReutersInstrumentCodeScheme
3. https://spec.edmcouncil.org/fibo/ontology/SEC/Securities/ParametricSchedules/ParametricSchedule

They were found by https://jenkins.edmcouncil.org/job/fibo/job/hygine_test_update/3

ElisaKendall:
@merelog The use of fibo-sec-sec-idind;ReutersInstrumentCodeScheme is strictly with respect to having an owl:sameAs statement in the deprecated resource, which should be allowed for named individuals

ElisaKendall:
@merelog The use of fibo-sec-sec-sch;ParametricSchedule is strictly with respect to having an editorial note that we have not removed, mainly due to some detail that we might want for further revisions downstream. I can remove the note, which would address the hygiene issue, but if you add something to ignore annotations on deprecated resources, then it would not show up as being still in use.

mereolog:
This looks fine, but note that I fixed the [check for the deprecated resources](https://github.com/edmcouncil/fibo/blob/hygine_test_update/etc/testing/hygiene_parameterized/testHygiene_resources_deprecated.sparql), and now it finds quite a lot of them — see: https://jenkins.edmcouncil.org/job/fibo/view/change-requests/job/PR-2026/3/.

I looked at a couple of them, and they seem to be genuine issues, e.g., https://spec.edmcouncil.org/fibo/ontology/FND/Accounting/ISO4217-CurrencyCodes/HRK, in the sense that they are deprecated resources that relate to other resources.

PR #2027, opened 2024-06-27
GitHub-2021 - Deprecated resources in use

ElisaKendall:
## Description

Eliminated remaining references to hasIdentity, which is deprecated. 

Fixes: #2021 

## Checklist:


jfgemski (approved):

ElisaKendall:
@mereolog I think the deprecated resource test should exclude reference data such as the MIC codes, currency codes and FpML rates, which are causing the new issues. This is because they are deprecated in the source reference data but there so that anyone depending on them with applications can still use them until they migrate. We want to represent these exactly as they occur in the reference sources, so the use of these resources, for example in codes and currencies, should not count as an issue. Perhaps the test should exclude individuals, and only apply to classes and properties?

mereolog:
> @mereolog I think the deprecated resource test should exclude reference data such as the MIC codes, currency codes and FpML rates, which are causing the new issues. This is because they are deprecated in the source reference data but there so that anyone depending on them with applications can still use them until they migrate. We want to represent these exactly as they occur in the reference sources, so the use of these resources, for example in codes and currencies, should not count as an issue. Perhaps the test should exclude individuals, and only apply to classes and properties?



> @mereolog I think the deprecated resource test should exclude reference data such as the MIC codes, currency codes and FpML rates, which are causing the new issues. This is because they are deprecated in the source reference data but there so that anyone depending on them with applications can still use them until they migrate. We want to represent these exactly as they occur in the reference sources, so the use of these resources, for example in codes and currencies, should not count as an issue. Perhaps the test should exclude individuals, and only apply to classes and properties?

Ok - I restricted the test and there is one more guy to look at: https://spec.edmcouncil.org/fibo/ontology/SEC/Securities/ParametricSchedules/ParametricSchedule

mereolog (approved):

jfgemski (approved):

Issue #2039, opened 2024-08-26
Implicit punning of OWL Classes in triples (as objects) with object property


mereolog:
There are 40 odd OWL classes which occupy the object position in triples where the predicate is an object property - see: 
[classes_in_triple_object_positions.xlsx](https://github.com/user-attachments/files/16744993/classes_in_triple_object_positions.xlsx)

This implies that these classes are also individuals.

PR #2049, opened 2024-09-18
GitHub-2039 - Implicit punning of OWL Classes in triples (as objects) with object property

ElisaKendall:
## Description

Revised the modeling strategy for ISO 15022/20022 corporate actions and example market indices baskets to eliminate punning issues

Fixes: #2039 

## Checklist:

### DIFF
--- a/CAE/CorporateEvents/ISO15022-CorporateActionIndividuals.rdf
+++ b/CAE/CorporateEvents/ISO15022-CorporateActionIndividuals.rdf
@@ -50,205 +50,305 @@
 		<owl:imports rdf:resource="https://www.omg.org/spec/Commons/Designators/"/>
 		<owl:versionIRI rdf:resource="https://spec.edmcouncil.org/fibo/ontology/CAE/CorporateEvents/ISO15022-CorporateActionIndividuals/"/>
 		<fibo-fnd-utl-av:hasMaturityLevel rdf:resource="&fibo-fnd-utl-av;Provisional"/>
-		<cmns-av:copyright>Copyright (c) 2016-2023 EDM Council, Inc.</cmns-av:copyright>
+		<cmns-av:copyright>Copyright (c) 2016-2024 EDM Council, Inc.</cmns-av:copyright>
 	</owl:Ontology>
 	
 	<owl:NamedIndividual rdf:about="&fibo-cae-ce-15022;BIDS">
 		<rdf:type rdf:resource="&fibo-cae-ce-act;BusinessStrategyClassifier"/>
+		<rdf:type>
+			<owl:Restriction>
+				<owl:onProperty rdf:resource="&cmns-cls;classifies"/>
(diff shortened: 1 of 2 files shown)


jfgemski (dismissed):

tahoeblue (dismissed):

rivettp (changes requested):
It is quite an unintuitive modeling approach to have restrictions on Individuals, and I would need to understand the rationale. 
Worse, it seems to have a flaw - use of someValuesFrom requires at least one instance to be classified by any classifier.

ElisaKendall:
@rivettp It is perfectly legal to do this, and we used this pattern quite successfully at Wells Fargo and on other projects such as IDMP. It is needed when an individual classifies something when it applies to all members of that class. We originally applied this approach for the CFI classification scheme, but agreed to change that given that every organization uses the CFI classification and it is now mandated by the FDTA. Many organizations do not use ISO 15022 to classify corporate actions, however, and we would be mandating its use in all cases if we follow the pattern used for the CFI. If people are using this ontology then they are committing to classifying corporate actions using the ISO 15022 classification scheme, and so the at least one corporate action classified by that scheme is entirely appropriate.

ElisaKendall:
@rivettp What we had previously caused punning, so either we do this, or we add restrictions on the elements of the two ontologies that represent the corporate actions in this local ontology. That's less intuitive IMHO.

rivettp:
You didn't address my second comment that it requires at least one instance of any action.

ElisaKendall:
@rivettp It does, but using an open world assumption one would say that there is an instance of such an action somewhere in the universe but we may not know about it. Also perfectly legal. It would only be required if you close the world.

ElisaKendall:
@rivettp The other option would be to change it to either all values from or min 0. I can make that change, but all values from might be problematic for people wanting to use the ACTUS classifiers.

ElisaKendall:
@rivettp I would have to try it with examples, using two classifiers in different schemes for the same thing, such as a repurchase offer, then run the reasoner to see what happens. It might make a logical equivalence between the two, but that might be ok if they are truly equivalent. With some values from, that leaves the door open for some actions to not be classified by a particular code.

rivettp:
The restriction would apply only to an individual code such as BONU and declares it's a member of the class of things that classify only instances of BonusIssue. 
I think. 
The fact that it's hard to think through, and even you feel the need to discover what happens in a reasoner, is a bit of a warning sign that we need to tread carefully with this pattern. I think something that's simple and obvious may be more important than an extra level of checking. Especially since there are many checks we forget about even trying due to reasoner performance, even for toy examples let alone enterprise data. Not to mention for DL safe reasons.
And we shouldn't forget about the possibility of expressing constraints using SHACL - which is what it's designed for.

rivettp (approved):
Happier with the allValuesFrom

jfgemski (approved):

Issue #2070, opened 2024-11-14
FND-386 - Eliminate deprecated elements that have been deprecated for more than 2 quarters


ElisaKendall:
This includes many of the elements deprecated with the move to Commons 1.1 almost a year ago, which are mapped in the FND area of the wiki.

PR #2071, opened 2024-11-14
FND-386 - Eliminate deprecated elements that have been deprecated for more than 2 quarters

ElisaKendall:
## Description

Eliminate long-time deprecated elements, particularly those elements deprecated due to integration of the OMG Commons Ontology Library v1.1, from the FIBO FND domain area. This update eliminates two ontologies, namely Roles and Quantities and Units, in favor of their replacements in Commons.

Fixes: #2070 / FND-386

## Checklist:


rivettp (approved):
The following are flags for future review while we have the record of the replaced elements (though not their definition).

Not happy about exemplifies generally, even less it replacing realizes.

Not sure the BusinessDates holdsDuring is really the same original meaning as its replacement in Parties and Situations.

Likewise hasRecuurenceStarDate seems more specific than hasStartDate.

There are several deleted deprecated classes in SecuritiesRestrictions with no replacement equivalentClass - was that deliberate?

PartyInRole seems distinct from PartyRole

We should check why Analytics had both Ratio and RatioValue, both replaced by Ratio

ElisaKendall:
> The following are flags for future review while we have the record of the replaced elements (though not their definition).
> 
> Not happy about exemplifies generally, even less it replacing realizes.
[efk} -- realizes has a different definition in Commons, so we should revisit but did not want two properties with the same name that might confuse users
> 
> Not sure the BusinessDates holdsDuring is really the same original meaning as its replacement in Parties and Situations.
[efk] Happy to discuss but I think it is.
> 
> Likewise hasRecuurenceStarDate seems more specific than hasStartDate.
[efk] Ok - we should think about this as we have use cases for it.
> 
> There are several deleted deprecated classes in SecuritiesRestrictions with no replacement equivalentClass - was that deliberate?
[efk] they were moved to a US-specific and/or EU specific jurisdiction ontology -- still there, just improved and made specific to the jurisdiction.
> 
> PartyInRole seems distinct from PartyRole
[efk] not the way we've used it to date
> 
> We should check why Analytics had both Ratio and RatioValue, both replaced by Ratio
[efk] only one was used and we reviewed that on a couple of Commons calls to make sure that it was the right approach, though you may not have been on.

jfgemski (approved):

